PR #5, opened 2026-05-27
fix: UX, runtime, and docs improvements for first-run experience

uzayer:
## Summary

- **Frozen input placeholder** — CodeMirror baked the placeholder string at mount time (empty `useEffect` deps). Wrapped it in a `Compartment` so it reconfigures when the phase changes from `checking` → `ready`.
- **Silent agent bind failure** — `ensureBound()` errors were swallowed by `console.warn`, leaving the input silently broken. Now captured in `bindError` state with a red "Couldn't connect to agent — Retry" pill above the input.
- **Checking phase label** — `disabledReason` had no case for `"checking"`, showing the generic fallback during the startup probe. Now shows "Checking Claude Code…".
- **fnm PATH support** — `apply_cheap_path_extras()` only enriched nvm paths. Users with fnm had `npx` only at `~/.local/share/fnm/aliases/default/bin`, causing `npx -y @agentclientprotocol/claude-agent-acp` to fail with ENOENT and surface as `acp: driver task panicked before initialize`.
- **`claude_status` timeout** — `claude --version` and `claude auth status` had no timeout. Under load the Tauri executor could starve the command indefinitely, freezing the phase at `"checking"`. Added an 8 s timeout.
- **README** — split prerequisites into "To run" vs "To build from source"; rewrote the Claude Code integration section to describe the actual ACP stack (stale text said "Atlas does not use ACP").

## Pre-existing issues (not introduced here)

- `npx tsc --noEmit` fails with one type error in `editor-panel.tsx` (`Cannot find module '@lezer/highlight'`) — confirmed pre-existing on `main`.
- `npm run lint` is broken — no `eslint.config.js`; ESLint v9 requires the flat config format.

## Test plan

- [ ] App loads — input briefly shows "Checking Claude Code…" then becomes enabled
- [ ] If the ACP bridge fails to spawn, the red retry pill appears and clicking it retries
- [ ] Sending a message works end-to-end
- [ ] Users with fnm as their Node version manager can start the app without ENOENT errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,14 @@ Concretely:
 
 ### Prerequisites
 
+**To run the app (production .dmg):**
+- **Claude Code CLI** — install per [Anthropic's instructions](https://docs.anthropic.com/en/docs/claude-code). Must be on your `PATH` (typically `~/.local/bin/claude`). Required for the AI agent — the app checks for it on every launch.
+- **Node.js 18+** — required at runtime to spawn the ACP bridge (`npx -y @agentclientprotocol/claude-agent-acp`). On first launch the bridge package is downloaded and cached; subsequent launches are instant. If you installed Node via `nvm`, make sure at least one version is active.
+
+**To build from source (additional):**
 - **Node.js 20+** and a package manager (`pnpm`, `npm`, or `bun` — `bun` is the one wired into `tauri.conf.json` by default).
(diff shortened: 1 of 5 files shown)


pacifio:
Already fixed with 0.1.10

PR #6, opened 2026-05-27
fix(landing): update domain to tryatlas.cc

uzayer:
Updates the stale `atlas.antarys.ai` domain to `https://www.tryatlas.cc/` in two places: the landing page README and `index.html`.


pacifio:
Already fixed with v0.1.10

PR #46, opened 2026-07-27
Checkpoints & Offline Capture (ATL-79)

ahammadnafiz:
# Checkpoints & Offline Capture

Implements all 12 sub-issues of [ATL-79](https://linear.app/tryatlas/issue/ATL-79). Records every agent Session locally, links each to the commits it produced, and — for Cloud Workspaces — syncs that record to the Organisation. Offline and account-free usage are first-class, not degraded modes.

## Tickets

- [x] **ATL-80 — atlas-redact: secret scrubbing crate**
  Six layers (Shannon entropy, 312 vendored betterleaks rules, provider prefixes, credentialed URIs, connection strings, bounded credential k/v) with JSON-aware traversal. Rule corpus is *generated* from upstream TOML, so a version bump is a re-run. Pure: no I/O, no async, no Tauri.

- [x] **ATL-81 — Capture Sessions and Messages into the local store**
  New `atlas-checkpoint` crate. `.atlas/sessions.db` in WAL, tables mirroring the eventual server shape, redact-on-write, titles derived from the first prompt, >64 KB bodies spilled content-addressed, per-Workspace writer lock, one-turn-one-transaction.

- [x] **ATL-82 — Record tool calls, file touches and edit patches**
  Tool calls as rows with a *derived* canonical name; both `apply.rs` wire-truth defects fixed; NFC path normalisation; `existed_before` sampled at write time; binary results stored verbatim.

- [x] **ATL-83 — Create Checkpoints from observed commits**
  The asymmetric link rule, per-Workspace commit cursor, first in-process watcher consumer, open-time backfill, and every commit shape (merge/initial/rename/delete/detached/cursor-loss).

- [x] **ATL-84 — Keep Checkpoints attached across rebase, amend and squash**
  git patch-id reconciliation. Amend and rebase re-point silently; squash orphans honestly; ambiguity prefers the recorded branch and otherwise orphans rather than guessing.

- [x] **ATL-85 — Workspace binding, Local mode, and the enable popover**
  Binding record + enable popover in the Workspace switcher. Local needs no account and no network. Identity signals are evidence, never gates.

- [x] **ATL-90 — Capture health: one visible signal**
  Per-Workspace OK/Degraded/Stopped with a reason *and* a next step. Watcher liveness read from the registry, never inferred from silence.

- [x] **ATL-91 — Import external agent transcripts**
  Backfill + ongoing per-Workspace scan of Claude Code JSONL. Cross-source dedupe, no Checkpoints for imported Sessions, resumable, malformed-line tolerant.

- [x] **ATL-86 — Cloud create: Slug availability and Organisation binding**
  Debounced availability with three answers (available/taken/**unknown**), server-first registration, advisory fingerprints.

- [x] **ATL-87 — Drain the outbox to ingest**
  Batched drain with count + byte ceilings, blob-first ordering, per-artifact results with batch-bisect fallback, mid-drain token refresh, 401/403/429 distinguished.

- [x] **ATL-88 — Connect an existing repository to a teammate's Workspace**
  Pre-selection as a pure function: one match selects, a fork selects via root commit, **several matches select nothing**.

- [x] **ATL-89 — Promote a Local Workspace to Cloud with history backfill**
  Promotion flips `local` → `pending` and reuses the existing drain. No separate backfill path.

## Implementation summary

Two new crates, both Tauri-free and testable without the runtime:

- **`atlas-redact`** — string in, redacted string out, plus a per-category tally driving the "N secrets redacted" figure.
- **`atlas-checkpoint`** — store, capture, git linkage, importer, outbox.

Redaction runs **before persistence, not before upload**, so the local store is never itself a disclosure risk and every later feature inherits the guarantee. Commits are **observed, not intercepted** — no hooks, no shadow branches, no commit trailers, no refs written — which is what lets a commit made from a terminal, another editor, or while Atlas was closed still find its Session.

## Findings that changed the implementation

Each was caught by a test rather than by review:

1. **Merge commits double-count Checkpoints.** First-parent *traversal* stops side commits being re-walked, but the merge commit's own first-parent diff still carries all that work. `git pull` produces merges constantly, so linking them roughly doubles Checkpoints on any branch that pulls. Merges now create none; the residual trade-off is documented at the call site.

2. **`.atlas/` needed its own `.gitignore`.** The existing `ensure_atlas_gitignore` is a user-toggleable setting that appends to the *project's* file. Tolerable for JSON; a SQLite database brings `-wal`/`-shm` sidecars that got swept into commits and broke `git checkout`.

3. **Patch-id candidates must span every ref.** A cherry-pick is by definition the same diff on two branches, so scanning only HEAD makes the collision invisible and the re-match confidently picks the one candidate it saw.

4. **`record_prompt` is not idempotent**, which duplicated the first turn when re-reading a grown transcript. The importer now records every line through an id-carrying path.

5. **Test fixtures had to be assembled at runtime.** Structurally-valid secrets are what makes a redaction test meaningful — and they trip push protection. Joining them from parts keeps the test's power without teaching anyone to click "allow this secret".

## Validation

- **Tests:** 319 passing — 95 `atlas-redact`, 205 `atlas-checkpoint`, 19 `atlas-agents` (incl. 4 pinning the `apply.rs` defects). Integration suites drive real temporary git repositories with real git commands, real SQLite databases, fixture transcript directories, and a real HTTP server on loopback. No mock traits.
- **Lint:** `cargo clippy --all-targets` clean on both new crates.
- **Typecheck:** `tsc --noEmit` clean.
- **Build:** `cargo check` clean on `src-tauri`.
- **Not run:** `npm run lint` fails on a clean checkout too (pre-existing ESLint flat-config migration), untouched here.

## Notes

**⚠️ Blocked on ATL-57 for the Cloud half.** Re-verified against the server: the ingest artifact union is still `agent_message | memory_fact | kb_item | git_ref`; `ROUTING_DB` is bound and never referenced (no dedupe); the queue consumer acks *before* the workflow runs (no DLQ); one bad org 403s the whole batch. ATL-86/87/88/89 are therefore built against the contract and verified with a loopback stub — `tests/drain.rs` is effectively the written form of that contract. **This must not ship against an endpoint that has not adopted it.**

**Deliberately not done**, each recorded on its Linear issue rather than dropped silently: Codex session-store import; the Connect and Promote UI dialogs (logic is complete and tested, the surfaces are not built); drain-progress rendering; ACP token-split backfill; and Attribution itself, which is out of scope for this spec — its inputs are captured.

**Local mode is complete and shippable today.** ATL-80→85, 90 and 91 have no server dependency: capture, commit linkage, rebase survival, the importer and the health signal all work with no account and no network.


ahammadnafiz:
@pacifio merge

PR #54, opened 2026-07-31
fix: use non interactive login shell for resolution

Fardays90:
Problem: 
Running bun run dev:app (or any dev mode launch via tauri-cli) would open the window, show the initial skeleton UI and then become completely unresponsive no errors. This was reproducible on Linux and is likely reachable on macOS too when running dev builds from an actual terminal (as opposed to a Finder/Dock launched production build).

Root cause:
claude_setup.rs::resolve_cli and node_setup.rs's mirrored copy both resolve a CLI's path by spawning the user's shell:
```rust
AsyncCommand::new(&shell)
    .args(["-lic", &format!("command -v {name} 2>/dev/null")])
    .output();
```
The -i (interactive) flag was unnecessary and is the actual bug. An interactive shell on startup runs job control initialization. It checks whether it's the terminal's current foreground process group and, if not, tries to claim it via tcsetpgrp(). That's fine when there's no controlling terminal at all (a Finder-launched GUI app the ioctl just fails with ENOTTY and the shell moves on silently). But when this probe is spawned from a real terminal session  or any dev-mode run via bun run dev:app,  the process inherits a real controlling terminal and tokio::process::Command spawns children into the parent's process group by default (no new pgroup, no pty). So the shell's job-control bootstrap collides with the terminal's actual foreground group and gets sent SIGTTOU. Default disposition: stop. Because it shares its process group with atlas, cargo, and everything above it in the tree, the stop signal hits the whole group at once which is why the entire app froze, not just one subprocess.

Fix:
Drop -i, keep -l

```rust
AsyncCommand::new(&shell)
    .args(["-lc", &format!("command -v {name} 2>/dev/null")])
    .output();
```
-l alone still sources login shell PATH configuration which is the only thing this lookup actually needs command -v doesn't require interactive shell behavior (job control, prompts). Non interactive shells skip job control init entirely so there's no terminal claim attempted and no SIGTTOU.

### DIFF
--- a/crates/atlas-acp/src/spawn.rs
+++ b/crates/atlas-acp/src/spawn.rs
@@ -90,7 +90,7 @@ fn resolve_program_abs(program: &str) -> Option<String> {
     resolved
 }
 
-/// Run `$SHELL -lic <script>` with an OWNED timeout: on expiry the probe child
+/// Run `$SHELL -lc <script>` with an OWNED timeout: on expiry the probe child
 /// is killed (so its reader thread exits promptly) instead of being abandoned
 /// to run forever — the old `recv_timeout`-only pattern leaked one thread AND
 /// one login shell per timed-out probe.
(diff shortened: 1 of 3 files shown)


pacifio:
didn't notice this, thanks for the pr!

PR #56, opened 2026-08-01
chore: issue + PR templates, dependabot, CONTRIBUTING pass

uzayer:
### What?

Adds the GitHub contribution surface the repo was missing: two issue forms, a PR template, and a Dependabot config. Also reworks the parts of `CONTRIBUTING.md` that were out of date, and points the in-app feedback panel at the new forms.

### Why?

Issues were arriving in wildly different shapes — some one-line notes, some full writeups, some with invented "Steps to Reproduce" sections that nobody had actually walked. A heavy template makes people fabricate the fields it demands; no template makes reports unroutable. These two forms require exactly one field (a freeform description) and make everything else optional, so a one-line note and a long investigation are both valid submissions.

`crates/*` packages aren't `src-tauri` workspace members, so Dependabot doesn't discover them recursively — each needs its own entry.

### How?

**Issue forms** — `bug_report.yml` (description · area · confidence · prior art) and `feedback.yml` (description · area · prior art). Only `description` is required on either. The optional fields are ones where leaving them blank is an honest answer: "Not sure" is a real option for area, and prior art ("Zed puts the model picker in the panel") is often more implementable than a paragraph of description.

**`config.yml`** — deliberately keeps blank issues **enabled**. Already-installed Atlas builds link to `/issues/new?title=…&body=…` with no `template=`; disabling blank issues redirects that to the chooser and silently drops the prefilled body, losing whatever the user typed in the feedback panel. The reasoning is recorded in a comment in the file so it doesn't get flipped back.

**PR template** — branch-targeting guidance up top, and a checklist trimmed to only what CI can't verify (`cargo check`, per-crate `cargo test`, running the app). CI already runs `bunx tsc --noEmit`, so there's no box for it.

**`feedback-links.ts`** — each feedback category now opens the matching form and prefills the `description` field. No dropdown prefill: matching an option by exact string is a cross-file contract nothing typechecks, and the title prefix already carries the category.

**`CONTRIBUTING.md`** — quickstart now shows the clone/install/run path and states that no API keys or `.env` are needed; fork/branch/PR and bug-reporting sections updated to match.

Note that issue templates only take effect once they reach the default branch, so none of this is user-visible until `0.2.5` merges to `main`.

### Verification

- `bunx tsc --noEmit` — clean
- All three issue templates parse as YAML; each declares exactly one required field
- Rebased onto current `upstream/0.2.5` (`da5dd5b`); no file overlap with the timeline/linux commits

Fixes #

## Checklist

- [x] `cd src-tauri && cargo check` passes — no Rust touched
- [x] `cargo test` passes from inside every crate you touched — no crates touched
- [x] You've run the app and used the change in a window — **not done.** The only runtime change is `issueUrl()`, and the forms it targets don't resolve on GitHub until this reaches the default branch, so the link can't be exercised end-to-end yet. Worth a manual click-through from the feedback panel after the release merge.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,81 @@
+name: Bug Report
+description: Something is broken, or behaves in a way it shouldn't
+labels: ["bug"]
+body:
+  - type: markdown
+    attributes:
+      value: >-
+        Write as much or as little as you have. A one-line note is a fine issue;
+        so is a long writeup with code references. Only the first box is required —
+        everything below it is optional, so skip anything that doesn't apply.
+
+  - type: textarea
+    id: description
+    attributes:
+      label: What's going on?
+      description: >-
+        Markdown and code blocks render here, so this box can hold anything from a
+        single sentence to a full investigation. If you have them, useful things to
+        include: what you expected instead, how to see it in the app, screenshots or
+        a screen recording, relevant log lines, what you've already checked or ruled
+        out, and your Atlas version / OS + version / which agent was active if it's
+        relevant.
+      placeholder: |
+        Terminal tab loses scrollback when I switch away and back.
+
+        Only happens after the pane has been split. Probably around
+        src/features/terminal/stores/ — the buffer lives in the PTY, so
+        the remount may be dropping the subscription rather than the bytes.
+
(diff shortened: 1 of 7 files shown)


uzayer:
clippy step failing was pre-existing on 0.2.5 for the timeline work. this PR did not touch anything related to the failing test

Issue #32, opened 2026-07-03
Add rendered Markdown preview in file viewer


uzayer:
# Problem

Atlas needs a first-class Markdown rendering surface for Markdown files/documents, rather than showing source-only content or relying on chat-specific rendering. Tables are particularly hard to use in the agent panel when content forces narrow columns.

# Existing implementation

The knowledge base already includes a suitable starting point:

- ```src/features/knowledge/components/readme-view.tsx``` renders repository README Markdown at document scale with GFM tables, syntax highlighting, raw HTML support, images, and ```<details>```.

- ```src/lib/markdown.tsx``` is a smaller shared renderer for chat/canvas/memory.

- ```src/lib/markdown-render.ts``` plus ```markdown-cache.tsx``` is the chat-optimized, sanitized worker pipeline.

These are currently separate renderers with duplicated component styling. The KB README renderer is the closest UX baseline for a document/file preview; the chat renderer must remain safe for untrusted agent output.

# Scope

Add a read-only Markdown preview mode/surface in Atlas, using the knowledge-base document renderer as the visual and capability baseline while extracting or sharing only the pieces that are safe for the target context.

Improve rendered-table usability in the same work:

- prevent content from being forced into unreadably tight columns;

- allow a table to use its intrinsic/minimum width and scroll horizontally within its container;

- treat manual column resizing as a progressive enhancement only if it can be implemented accessibly and persisted/safely within the rendered view. It is not required for the initial delivery.

# Acceptance criteria

- Markdown files/documents can be viewed as rendered Markdown from the relevant Atlas surface, with an obvious way to return to source/editing where applicable.

- Rendering supports the established KB baseline: GFM tables/task lists, code highlighting, links, images, and common README HTML such as details.

- Rendering policy is explicit: untrusted agent content is sanitized; any raw-HTML allowance is limited to trusted/local document contexts.

- Long/wide tables retain readable cell widths and provide horizontal scrolling rather than collapsing columns.

- Do not add drag-to-resize table columns unless the implementation is low-complexity, accessible, and does not create per-document state-management debt.

- Add representative renderer/table regression coverage.

PR #58, opened 2026-08-01
fix: markdown preview for fileviewer

Fardays90:
#32  

**changes**

1.  Added a dedicated `MarkdownFile` component in `src/lib/` for Markdown document previews.
2. Added an explicit trusted prop to control raw HTML rendering.
3. Trusted/local documents will render perfectly.
4. Untrusted content can be rendered without enabling raw HTML.

**Improved table rendering**

1. Tables now use their intrinsic width (min-w-max).
2. Wide tables scroll horizontally within their container.
3. Prevents content from collapsing into unreadably narrow columns.
4. Integrated the preview into the Atlas editor by rendering Markdown when Preview mode is selected.

pr1
scrollable tables


pacifio:
This was very much needed, thanks for the good work! @Fardays90

Issue #78, opened 2026-08-02
Focus the terminal input reliably after Cmd+J


uzayer:
## Problem

Cmd+J opens or reveals the terminal tab but does not reliably put the text cursor in the terminal. Users should be able to press Cmd+J and type immediately.

## Investigation

* `src/App.tsx` activates/creates the terminal then dispatches `atlas:focus-terminal` after one `requestAnimationFrame`.
* `BlockTerminal` listens for that event but ignores it until `isActive` is true, focusing either its command input or xterm alt-screen surface.
* On a newly mounted terminal or a tab becoming visible, the one-frame dispatch can arrive before the component is active/listening, losing the focus handoff.

## Acceptance criteria

* Cmd+J always focuses the intended terminal input (or active xterm alt-screen surface) after opening or revealing a terminal, including first-open, existing hidden terminal, split-view, and workspace-switch cases.
* No mouse click is required before typing.
* Focus behavior does not steal focus when Cmd+J is used to toggle away from the terminal.
* Use a deterministic readiness/acknowledgement or pending-focus contract instead of timing-only retries.
* Preserve the active terminal pane selection and accessibility focus semantics.

PR #87, opened 2026-08-03
fix: determinstic focus mechanism for terminals instead of race

Fardays90:
#78 

A deterministic terminal focus handoff for `Cmd+J` so the cursor reliably lands in the intended terminal input after opening or revealing a terminal. The change replaces the previous one frame focus event with a pending focus contract stored in the terminal store. Then lets each terminal instance claim focus only when its tab is active and its surface is ready. This makes focus behavior reliable across first open, hidden terminal, split view, and workspace switch scenarios while avoiding focus steals when toggling away from the terminal.


pacifio:
Ah this was really needed, thanks @Fardays90

PR #92, opened 2026-08-04
feat: add stop control for running foreground processes with escalation to force quit

TriptoAfsin:
feat: add stop control for running foreground processes with escalation to force quit)


pacifio:
This was already resolved in a previous PR that was merged so closing this now

PR #97, opened 2026-08-05
fix(atlas-redact): collapse identical if/else-if branches (clippy)

ithun-y-ittesaf:
There's a function that splits keys like myAPIKey into words (my, api, key). It had two separate if checks for "is this the start of a new word?" and when either one was true, they both did the exact same thing. Rust's linter (clippy) flags that as a mistake ("why write it twice?"), and CI treats linter warnings as hard failures, so the build was red.

Fix: combined the two checks into one if (with an "or") since they led to the same action anyway. No behavior change, verified against the existing 65+ test suite.

### DIFF
--- a/crates/atlas-redact/src/credential.rs
+++ b/crates/atlas-redact/src/credential.rs
@@ -178,9 +178,13 @@ fn segments(key: &str) -> Vec<String> {
             continue;
         }
         let next_lower = chars.peek().is_some_and(|nc| nc.is_ascii_lowercase());
-        if ch.is_ascii_uppercase() && prev_lower && !current.is_empty() {
-            out.push(std::mem::take(&mut current));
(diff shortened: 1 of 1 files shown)


ahammadnafiz:
@ithun-y-ittesaf can you check branch 0.2.5 is this issue solved on this branch because i have updated the test cases

ithun-y-ittesaf:
> @ithun-y-ittesaf can you check branch 0.2.5 is this issue solved on this branch because i have updated the test cases

Sure, let me get back to this.

uzayer:
Thanks for the fix! This same clippy issue (identical if/else-if branches in the acronym-splitting logic) was independently fixed in 84e7204 ("fix(atlas-redact): merge identical if/else-if branches in acronym split"), already merged to main/0.2.6 — same collapsed condition. Closing as a duplicate.

Issue #64, opened 2026-08-02
Make Atlas settings config-driven and add an agent self-configuration skill


uzayer:
## Goal

Every user-facing Atlas setting should be inspectable and changeable through a documented configuration file, with the Settings UI and config file remaining in sync. Ship an Atlas-managed skill that lets supported agents discover and safely update that configuration—analogous to Claude Code’s self-update DX.

## Current state

* Global UI settings already have a typed `AppSettings` schema on both sides:
  * `src-tauri/src/state/app_state.rs`
  * `src/features/project/stores/project-store.ts`
* Those settings are persisted inside the app-data `state.json`, through a frontend-built app-state payload.
* Some adjacent configuration is fragmented into dedicated app-config files (for example `mcp-servers.json`, telemetry config, BYOK storage, and per-project `.atlas/` state).
* The existing Skills feature has a canonical managed library and can project a global skill into detected agent config folders (Claude Code, Codex, etc.).

## Scope

1. Define a stable, human-editable Atlas configuration format and location for user-facing settings, including schema versioning, documented defaults, validation, and migrations.
2. Make UI changes and config-file edits converge on the same source of truth. External edits must be detected/reloaded safely; malformed or unknown values must not overwrite a working configuration.
3. Add a globally projectable, Atlas-owned **self-configure** skill that teaches agents how to:
   * locate and inspect the supported Atlas config;
   * explain valid keys and values;
   * make minimal, validated changes through the supported mechanism;
   * preserve comments/unknown fields where the chosen format supports them; and
   * distinguish user preferences from secrets, credentials, session history, and machine-managed state.

## Non-goals

* Do not expose API keys, auth/session tokens, telemetry identities, or arbitrary internal runtime/state files as editable “settings”.
* Do not replace per-project document/editor/session state merely because it is file-backed; establish explicit ownership and scope first.
* Do not give the skill permission to update Atlas binaries, packages, or arbitrary files. “Self-update” here means self-configuration.

## Acceptance criteria

* A documented config file covers all currently user-facing Atlas preferences, including the existing `AppSettings` fields and settings added later through a single extension path.
* Atlas loads it at startup, validates it, and surfaces actionable errors without losing the last known-good settings.
* Changes made in Settings update the config atomically; supported external file edits are reflected in the app without a restart where practical (otherwise clearly signal reload/restart).
* Migration from the current `state.json` settings is lossless and idempotent.
[...]

Issue #38, opened 2026-07-03
Set up Husky and conventional commit enforcement


uzayer:
**Description**
The repo should enforce a more consistent commit workflow with Husky and conventional commit checks. This would improve hygiene and make collaboration history more predictable.

**Expected Behavior**
The repository should use Husky hooks and conventional commit enforcement so commit quality checks run automatically before changes land.

**Environment**
`pacifio/atlas` repository tooling.

PR #98, opened 2026-08-05
fix(windows): unbreak src-tauri build on Windows/MSVC

ithun-y-ittesaf:
## Description:

First clean Windows/MSVC build attempt on this repo surfaced three independent bugs blocking compilation, plus a Tauri capabilities schema Windows needs:

1. `browser.rs`: called `.visible(false)` on _tauri::webview::WebviewBuilder_, which has no such method (that's WebviewWindowBuilder/WindowBuilder only). Failed to compile on every platform, not just Windows. Swapped for .hide() on the built Webview handle, preserving the "initially hidden" contract callers rely on.

2. `updater.rs`: unconditionally imported `std::os::unix::fs::FileExt`, but the module is a macOS-only DMG updater (mounts DMGs, shells out to `hdiutil/codesign/spctl/ditto`) that was never gated behind #[cfg(target_os = "macos")]. Split into updater_macos.rs (unchanged logic) and updater_stub.rs (no-op, same public API) so lib.rs and the frontend's invoke() surface stay platform-agnostic.

3. atlas-embed: tokenizers' default esaxx_fast feature builds esaxx-rs's C++ suffix-array backend with a hardcoded static CRT, which conflicts with the rest of the binary's dynamic CRT (via usearch/cxx) and fails MSVC linking with LNK2038. Disabled it — it only speeds up training a Unigram tokenizer; loading/running pretrained ones is unaffected.

4. Added src-tauri/gen/schemas/windows-schema.json, the generated Tauri capabilities schema for the Windows target.

5. No behavior changes on macOS/Linux.


uzayer:
Thanks for this — and for the writeup. Three genuinely independent bugs, each diagnosed down to the actual mechanism rather than worked around, which made this a pleasure to review.

A few notes on what changed on the way in:

**Rebased onto `0.3.1`** and retargeted the PR there — Atlas uses version branches, so feature PRs land on the current version branch rather than `main` (see CONTRIBUTING.md). The branch was ~349 commits behind, so a couple of things shifted:

- The `browser.rs` fix landed independently in the meantime (`062f5cd2`), so only your explanatory comment survives the rebase. Your diagnosis was right, it just got there twice.
- `src-tauri/Cargo.lock` no longer exists — #38 moved everything to one root workspace. Git followed the rename correctly and your `esaxx-rs` edit applied cleanly to the root lock. I confirmed with `cargo metadata --locked` that it is exactly what cargo would regenerate.
- `updater_macos.rs` in the PR was a snapshot of `updater.rs` from before the dual-arch update URIs (`uri_mac_arm`/`uri_mac_intel`) and the config.toml migration (#64) landed. Taking it as-is would have quietly reverted both, so I rebuilt it from the current `updater.rs` — the split is still a pure move with zero logic change.

**One follow-up commit** on top of yours. Declaring the four commands in both platform files tripped `tests/ipc-contract.test.ts`: it reads source text and cannot evaluate `#[cfg]`, so two mutually exclusive declarations look exactly like a real handler collision — and Tauri does key its IPC router on the bare fn name, so the guard is right to be strict. The signatures had also drifted (the stub's `update_ignore` was a sync `fn(version)` against the real `async fn(version, app)`), which nothing could catch since only one side compiles per target.

So the `#[tauri::command]` wrappers and both DTOs now live once in `updater.rs`, with the platform modules exposing `pub(super)` fns behind identical signatures — which also restores the "IPC verbs grouped into a single `commands/<domain>.rs`" convention. Mismatches now fail to compile instead of drifting.

**Verification:** `cargo check -p atlas` passes on the macOS path, and again with the dispatcher forced to the stub path to exercise the Windows half — worth doing since there is no Windows runner in CI. Your `esaxx_fast` reasoning checks out precisely: it is `default` minus exactly that feature, its only consumer is the Unigram trainer, and `esaxx_rs::suffix_rs` is unconditional, while atlas-embed only ever calls `Tokenizer::from_file`. Provably behaviour-neutral.

The one red CI check is `ci-coverage`/`atlas-comms`, which fails on a clean `0.3.1` too and is being handled in #224. Not yours.

Merging now. Thanks again — the updater being ungated was a real latent bug that just happened to only bite on Windows first.

Issue #63, opened 2026-08-02
Inherit user-level permission mode when creating agent sessions


uzayer:
## Problem

New Atlas sessions do not reliably start with the permission mode configured for the corresponding CLI agent at user scope.

Examples:

* A Claude Code user config that starts sessions in **Bypass Permissions** should result in a new Atlas Claude session starting in `bypassPermissions`.
* A Codex user config that starts the CLI in YOLO / full-access mode should result in a new Atlas Codex session starting in that equivalent advertised ACP mode.

## Investigation

Atlas already launches the ACP bridges in the host environment, so they can read the normal `~/.claude` / `~/.codex` configuration. The UI then overrides or adopts modes at the wrong points:

* `src/features/chat/stores/chat-store.ts` creates Claude sessions with `claudePermissionMode: "default"`.
* `src/features/chat/components/chat-panel.tsx` applies that tab mode before exposing the binding, but only sends `setMode` when it is non-default. Atlas never resolves a configured initial Claude permission mode first.
* Codex modes are optimistically restored from an Atlas ACP-mode cache and then populated from `agents.snapshot()` **after** `session/new`; this is not a user-config-derived session default and is too late to guarantee the first turn runs with the configured mode.
* The backend already has generic ACP mode setting via `agents_set_mode` / `Registry::set_session_mode`, and the native agent normalizes aliases including `bypassPermissions`, `danger-full-access`, and `yolo`.

## Scope

Make new sessions preserve the agent's user-level global permission preference as their initial mode, for every agent Atlas supports that exposes permission modes. Do not persist an Atlas-only default that overrides a CLI configuration unless the user explicitly changes the mode in Atlas.

## Acceptance criteria

* A newly created Claude Code session honors the user-level configured bypass/ask/edit/plan default before its first prompt can be sent.
* A newly created Codex session honors the user-level configured permission default (including full-access/YOLO equivalents) before its first prompt can be sent.
* The implementation is agent-agnostic where ACP exposes modes, validates selected ids against the agent's advertised modes, and does not leak a mode between agent types.
* An explicit mode change made in Atlas still takes precedence for that session.
* Add focused tests covering Claude and Codex initial-mode resolution, including the first-prompt ordering guarantee.

uzayer:
fixed in 0.3.0.x

PR #157, opened 2026-08-12
fix(agents): inherit configured permission modes

abubakarsiddik31:
## Summary
- preserve the mode returned by an agent's session creation instead of overwriting it with Atlas defaults
- apply an explicit Atlas mode only when the user selected one, before the binding can release the first prompt
- validate requested mode ids against the session's advertised modes while retaining compatibility with legacy agents

## Why
New Claude Code and Codex sessions could ignore the user-level permission configuration because Atlas initialized mode state independently of the ACP session response.

## Validation
- cargo check --manifest-path src-tauri/Cargo.toml -p atlas-agents
- cargo check --manifest-path src-tauri/Cargo.toml -p atlas
- bun run typecheck
- bun run lint
- bun run format:check
- bun run test -- --run (493 passed)

Note: the standalone atlas-agents unit-test invocation is currently blocked in this environment because its independent Cargo resolution requires an unavailable crates.io package (toml_write). The full Tauri compilation above succeeds.

### DIFF
--- a/crates/atlas-agents/src/error.rs
+++ b/crates/atlas-agents/src/error.rs
@@ -12,6 +12,8 @@ pub enum Error {
     SessionExists,
     #[error("session worker is gone")]
     WorkerGone,
+    #[error("mode is not advertised by this session: {0}")]
+    InvalidMode(String),
     #[error("acp: {0}")]
     Acp(#[from] atlas_acp::AcpError),
     #[error("io: {0}")]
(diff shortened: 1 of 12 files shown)


pacifio:
This I think only works for claude code, this is a good feature to have do you think you can make it work for all the acp agents, it should be easy to persist like `acp_agent_name:last_permission` or something, the latest work is happening inside `0.3.0-x`

abubakarsiddik31:
Good points — both done in the latest push.

The branch is now rebased on 0.3.0-x, and explicit mode picks are persisted per agent as atlas:last-mode:v1:<agentType>. Since registry agents use their plugin id as the agent type, every ACP agent gets its own record — so switching to full-access on Codex (or any installed agent) sticks across restarts.

Two notes on how I scoped it, per #63: only picks the user explicitly makes get saved — never modes the agent itself reports or switches into — and picking default clears the record instead of storing it, so the CLI's own config always wins unless the user says otherwise.

Small correction though: the original change wasn't Claude-only. Non-Claude agents already went through the generic ACP mode path (explicit override + validation against advertised modes) — Claude just needed extra plumbing since its mode shows up in a Claude-specific UI. The missing piece was exactly what you pointed at: persistence.

pacifio:
Ahh thank you for the correction, getting this merged right away!

PR #159, opened 2026-08-14
fix(agents): preserve late assistant output after turn completion

### LINKED ISSUE #94: Assistant messages from late-arriving/background turns are silently dropped after the turn finalizes
## Summary

When an agent turn finalizes (goes idle) and a *later* ACP notification for that same session arrives — e.g. an assistant message chunk produced by a detached/background subprocess that finishes well after the main turn ended — that notification is silently dropped by `atlas-agents`. It never reaches the canonical session message log and is never broadcast to the frontend, so the user never sees it. The message does exist in the underlying agent's own on-disk transcript (e.g. Claude Code's `~/.claude/projects/<slug>/<id>.jsonl`), just not in Atlas's state or UI.

## Reproduction

1. Open a chat session using the Claude Code agent backend in Atlas.
2. Ask it to kick off a long-running background/detached job (e.g. launch a Codex CLI review task that runs asynchronously and is expected to report back later).
3. The main turn finalizes normally almost immediately, since the "launch the background job" tool call reports itself `completed` as soon as the subprocess is spawned — there is no ACP tool call left open/pending for Atlas to treat as "still in flight."
4. Minutes later, the background job finishes. The Claude Code subprocess autonomously pushes a fresh `agent_message_chunk` / `session/update` notification into the same session summarizing the result — with no corresponding new `session/prompt` call from Atlas.
5. **Expected:** the summary message appears in the chat as a new assistant message.
6. **Actual:** nothing appears. The composer just sits idle. The user has to send a new message (e.g. "still running?") before anything changes — sending a new message starts a fresh turn, which happens to unstick the pipe, but the original background-result message is gone from the UI/state entirely (it's still recoverable only by reading the agent's own raw transcript file, not through Atlas).

## Root cause

This is a turn-identity/epoch bug spanning `atlas-acp` and `atlas-agents`, not a frontend rendering issue.

1. **`crates/atlas-acp/src/driver.rs:373-393`** — every inbound ACP `SessionNotification` is stamped with `turn = guard.current_turn()`, i.e. whatever turn epoch happens to be "current" *at the moment the notification arrives*. There is no correlation between the notification and the specific `session/prompt` call that logically produced it.

2. **`crates/atlas-agents/src/actor.rs:634-636`** (`finalize()`) — when a turn finalizes, `self.running` is set to `None`, but `self.current_epoch` is **never reset**. It keeps holding the now-retired epoch value indefinitely, until a brand new `Send` calls `mark_turn_started` again.

3. **`crates/atlas-agents/src/actor.rs:474-497`** (`handle_stream`, `ActorMsg::Acp` arm) — the turn-identity gate only applies a stamped event if:
   ```rust
   let live = self.running.is_some() && stamp == epoch;
   ```
   Once a turn has finalized, `self.running` is `None`, so `live` is unconditionally `false` for *any* subsequently arriving stamped event — regardless of whether the epoch itself is still "current." The event is dropped with only a `tracing::debug!("dropping event from dead turn")` log line (line ~487-493); it never reaches `apply_event`, so it's never written into `SessionState.messages` and never emitted as a `SessionDelta` on the `atlas:agents` Tauri channel.

This gate is intentional and correct for its original purpose — rejecting stragglers from a *cancelled or superseded* turn (the rapid Stop + re-send race), see the existing tests `late_event_from_old_turn_is_dropped` (`actor.rs:1065-1092`) and `late_tool_event_after_idle_does_not_spin` (`actor.rs:1197-1218`). The bug is that it *also* rejects legitimate new content that arrives after a turn finalized normally (not cancelled, not superseded) simply because nothing is "running" anymore — which is exactly the situation for any assistant output produced by work that continues asynchronously in the background past the point where Atlas considered the turn done.

## Why background/detached jobs trigger this specifically

Atlas only defers turn finalization while there's an ACP tool call in `Pending`/`Running` status (`crates/atlas-agents/src/session.rs:225-238`, `has_inflight_tool_calls`, used in `actor.rs:518-543`). A tool call that merely *launches* a detached subprocess plausibly reports itself `completed` as soon as the launch succeeds — the actual multi-minute background run isn't represented as an open ACP tool call at all, so there's nothing to keep the turn open for. This is presumably correct behavior for the *launch* step, but it means the eventual background result has no live turn to land in.

## Suggested fix direction

Not proposing a specific patch here, but two directions worth considering:
- Clear `self.current_epoch` (set to `None`) in `finalize()`, so post-finalize events fail open per the gate's own existing rule (`current_epoch == None` → pass through, per the comment at `actor.rs:481-482`). This alone would stop the silent drop, but the event would then need a sensible place to land — likely needs to be treated as a fresh, unsolicited assistant message appended to the session (not assumed to belong to any specific live turn).
- Alternatively, give background/detached tool calls a way to keep the turn (or a lightweight placeholder) open until they actually resolve, so the normal in-flight-tool-call deferral in `finalize()` naturally covers this case instead of a special post-finalize path.

Either way, this should be validated against the existing turn-identity tests (`actor.rs:1065-1092`, `actor.rs:1197-1218`) to make sure the legitimate straggler-rejection behavior isn't reintroduced as a regression.

## Impact

Silent, hard-to-notice data loss: the user believes the agent went idle/did nothing, when in fact it produced a substantive reply that's simply gone from Atlas's chat history (though still present in the underlying agent's own transcript file, if one exists). No error is surfaced to the user or logged above `debug` level.

NgoQuocViet2001:
### What?

Preserve assistant output that arrives after a normally completed turn, while keeping stale-event protection for cancelled and failed turns.

### Why?

ACP notifications can arrive after detached/background work completes. Once the prompt future finalized, the actor remained pinned to the retired epoch and dropped those notifications before transcript application.

### How?

Clear `current_epoch` only when a turn finishes successfully for a non-cancellation reason. Update the actor regression to expect a late assistant chunk to reach the session transcript; the existing cancellation/resend and late-tool protections remain green.

Fixes #94

## Validation

- `cargo test` in `crates/atlas-agents` (21 passed)
- `cargo check` in `crates/atlas-agents`

## Checklist

- [ ] You've run the app and used the change in a window

### DIFF
--- a/crates/atlas-agents/src/actor.rs
+++ b/crates/atlas-agents/src/actor.rs
@@ -162,8 +162,9 @@ pub struct SessionActor {
     /// The live turn's epoch, as returned by `mark_turn_started` — the value
     /// the backend stamps onto this turn's events. Inbound stamped events
     /// whose stamp doesn't match are stragglers from a superseded/cancelled
-    /// turn and are dropped. `None` = no gating basis (no turn started yet,
-    /// or `mark_turn_started` failed) → stamped events fail open (applied).
+    /// turn and are dropped. `None` = no live gating basis (no turn started,
+    /// a normal turn finished, or `mark_turn_started` failed) → stamped events
+    /// fail open (applied).
     current_epoch: Option<u64>,
     /// Sends that arrived while a turn was live. Supersede = cancel-then-send
     /// (the Zed pattern): the incoming Send cancels the running turn and
@@ -648,6 +649,12 @@ impl SessionActor {
         let (status, delta, sweep_to) = match result {
             Ok(stop_reason) => {
                 let cancelled = stop_reason == "cancelled";
+                // Normal completion may be followed by output from detached
+                // work. With no live epoch, stamped session traffic fails open;
+                // cancelled and failed turns retain their stale-event guard.
+                if !cancelled {
+                    self.current_epoch = None;
+                }
                 (
                     SessionStatus::Idle,
                     SessionDelta::TurnFinished { stop_reason, turn_seq },
@@ -1063,9 +1070,10 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn late_event_from_old_turn_is_dropped() {
-        // An event stamped with a finished turn's epoch must not be applied
-        // after that turn's terminal.
+    async fn late_assistant_event_after_normal_turn_is_applied() {
+        // A normal turn may launch detached work that reports back after the
+        // prompt future resolves. Its assistant output still belongs in the
+        // session transcript even though the actor is already idle.
         let (handle, sink, mut gates, _conn) = setup_gated(1);
         handle.control_tx.send(Control::Send("hi".into())).unwrap();
         settle().await;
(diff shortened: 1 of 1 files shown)


pacifio:
Ahh this is super helpful, thanks @NgoQuocViet2001

NgoQuocViet2001:
@pacifio Glad it helped — thanks for the quick review and merge!

PR #196, opened 2026-08-24
Esc clears the filter inputs in Settings tables

### LINKED ISSUE #165: Escape clears the filter inputs in Settings tables
> **Getting started:** read `CONTRIBUTING.md`, run `bun install`, then `bun run dev:app` (Tauri dev build). Rust commands need `export SDKROOT="$(xcrun --show-sdk-path)"` on macOS, and crates are path-linked (no workspace) — run `cargo` from inside the crate directory. `bun run test` + `bun run typecheck` must stay green.

`area:settings` · `area:ux` · ~30 min

The API Keys and Local Models tables each have a search box
(`src/features/settings/components/providers-settings.tsx`,
`src/features/settings/components/models-manager.tsx`). Pressing <kbd>Esc</kbd>
while focused should clear the query (and keep focus); today it does nothing.

**Do:** add an `onKeyDown` handler to both inputs.

**Done when:** Esc empties a non-empty filter in both tables; an already-empty
filter lets Esc propagate (so it can close the settings tab as before).

---

aunoshuarodosyaurpa:
## Summary
- Add an `onKeyDown` handler to the API Keys search input and the Local Models filter input
- Pressing Esc while the input has a non-empty query clears it and stops propagation (keeping focus on the input)
- An already-empty filter lets Esc propagate unchanged, so it still closes the settings tab as before

Closes #165

## Test plan
- [x] `bun run lint`
- [x] `bun run format:check`
- [x] `bunx tsc --noEmit`
- [ ] Manually verify in `bun run dev:app`: type in each search box, press Esc, confirm it clears and keeps focus; press Esc again on empty input and confirm settings tab closes

### DIFF
--- a/src/features/settings/components/models-manager.tsx
+++ b/src/features/settings/components/models-manager.tsx
@@ -91,6 +91,12 @@ export function ModelsManager() {
           <input
             value={query}
             onChange={(e) => setQuery(e.target.value)}
+            onKeyDown={(e) => {
+              if (e.key === "Escape" && query) {
+                e.stopPropagation();
+                setQuery("");
+              }
+            }}
             placeholder="Filter models…"
             className={cn(
               "w-full h-7 pl-8 pr-2.5 rounded-md bg-bg-elevated border border-border-default",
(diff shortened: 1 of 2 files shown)


aunoshuarodosyaurpa:
Closing — reopening from the correct account.

PR #216, opened 2026-08-31
atlas-agent-delta: quadratic projection, four transcript misreports, three retained tables

ahammadnafiz:
### What?

The seven `crates/atlas-agent-delta` findings from ATL-212: ATL-219, ATL-220,
ATL-221, ATL-222, ATL-223, ATL-224 and the ATL-225 rollup.

Two of them (ATL-219, ATL-221) could not be fixed inside the crate, so this also
touches `atlas-terminal` and `atlas-acp-thread`. One (ATL-220) is frontend-only —
the Rust side was already correct.

### Why?

Two performance defects, three misreports to the user, and three tables that
grew for the life of the process.

**ATL-219 / ATL-223 — the same defect in two files.** Both rebuilt an entire
accumulated string on every update in order to work out what had been added to
its end. For a terminal that meant decoding the whole buffer, rebuilding the
whole result and comparing it byte for byte against the previous copy, once per
output chunk, with the session's mutex held — so a chatty command stalled the
whole session, not just its own card.

Measured over 5x the output, before and after:

| | before | after |
| -- | -- | -- |
| terminal projection | 16.3x the time | 4.0x |
| assistant streaming | 15.5x the time | 3.7x |

Linear is 5x. The benchmarks are kept as `#[ignore]`d tests rather than deleted,
because both tickets' acceptance criteria are about how cost *scales*, and a
benchmark that no longer exists cannot be re-run against a later change.

**ATL-220 / ATL-221 / ATL-222 / ATL-224 — four ways a transcript lied.** Every
tool call in a reopened conversation rendered as succeeded, failed ones
included. Every message in it was stamped with the moment the snapshot was
taken, so real pauses collapsed and turn durations read as zero. A plan the
agent had abandoned stayed on screen. An image-only assistant turn showed a
blank bubble that vanished on reload.

**ATL-225 — three tables with no removal path.** Permission and elicitation
routes, and the pre-registered event stream of a session whose load failed.

### How?

Per commit. Three things are worth reading the diff for:

**The projection asks for a tail, it does not compute one.** `OutputBuffer` now
decodes on write and hands out `&str`, and `AcpTerminal::output_appended`
answers "what arrived since this byte offset" without touching the rest of the
buffer. It returns `None` — meaning *re-read everything* — for every case where
an offset from an earlier read stopped naming the same bytes: either buffer
dropped its front, output shrank, offset landed off a character boundary, or the
tail could only be served by copying the PTY half too. The cheap path is an
optimisation with a guard, not a new contract.

**ATL-221 is fixed as far as the protocol allows, and no further.** Entries are
stamped when the thread learns of them, which makes snapshots stable and keeps
every gap the app witnessed. It does **not** recover times for a thread rebuilt
by an agent's `session/load` replay — ACP carries no per-message timestamp, so a
replayed conversation is stamped at replay time. The ticket suggested a null the
UI renders as "unknown"; I did not do that, because `replaceMessages` already
substitutes `Date.now()` for a missing timestamp, so a null would be the same
fabrication one layer up. Making it honest means teaching four render sites to
show "unknown", which is a UI change with no ticket behind it. The limit is
written on the field instead of hidden.

**ATL-224 was a `needs-decision`.** Option A: the stream now skips an empty run
to match the snapshot. The slot is still mirrored, so run indices stay aligned
with the thread's and the run is announced the moment it has text. Option B —
carrying the image through as structured content, the only choice that actually
*shows* the user what the agent sent — is a wire-schema change and deserves its
own ticket.

#### One thing found on the way that is bigger than the ticket it sits in

ATL-225 asks for the permission and elicitation maps to be pruned "in `attach`'s
cleanup, which today removes only from `sessions`". That cleanup never runs. The
projection holds the only strong handle on the thread (the connection keeps a
`Weak`, the host drops its clone) and the thread owns the event sender — so
`events.recv().await` can never return `None`, the task never ends, and the
removal is unreachable.

So the leak was not two maps of route keys. It was every session's projection,
its thread, and its terminals with their live PTY handles, retained until Atlas
exits. And ATL-225's suggested fix would have been inert behind the same dead
path.

Closing is now told rather than noticed: `DeltaProjector::close_session`, called
from the host's existing close path, which is also what finally drops the
terminals whose `Drop` kills a running PTY (ATL-214's fix, which had nothing to
trigger it). The stream-end path stays as a backstop. **This deserves its own
ticket** rather than being buried in a Low rollup, and it is not covered by
ATL-225's title.

Fixes ATL-219, ATL-220, ATL-221, ATL-222, ATL-223, ATL-224, ATL-225

## Checklist

- [ ] You've run the app and used the change in a window

Every fix was falsified: reverted, its test confirmed failing, restored. Baseline
for `atlas-agent-delta` goes from 19 tests to 31 plus two benchmarks;
`atlas-terminal` 15 to 17; frontend gains `snapshot-message.test.ts`.

The unchecked box is honest. Two paths want a real window before this ships:
resuming a session that contains a failed tool call (ATL-220 — the card should
now render as failed, and its id should match the agent's), and watching a long
command stream into a tool card (ATL-219 — output should still arrive
incrementally and in order, and should still show the truncation marker once the
buffer starts dropping its front).

### DIFF
--- a/crates/atlas-acp-thread/src/connection.rs
+++ b/crates/atlas-acp-thread/src/connection.rs
@@ -349,6 +349,16 @@ impl dyn AgentConnection {
     }
 }
 
+/// Rewinding a session's history.
+///
+/// No implementation and no caller today — `remove_entries_from` is reached
+/// only by the native agent's `/undo`. Whoever wires this up owns a coupling
+/// that is not local to this file: shrinking `AcpThread::entries` makes
+/// `atlas-agent-delta`'s per-session mirror shrink with it, and anything keyed
+/// by entry position or by tool call id (the projector's `open_permissions`,
+/// its `Projected` mirror) has to be trimmed in the same step. The projector
+/// handles the `EntriesRemoved` event it already receives; check it still does
+/// before assuming a new truncation path is free.
 pub trait AgentSessionTruncate: Send + Sync {
     fn run(&self, client_user_message_id: ClientUserMessageId) -> BoxFuture<'static, Result<()>>;
 }
--- a/crates/atlas-acp-thread/src/terminal.rs
+++ b/crates/atlas-acp-thread/src/terminal.rs
@@ -212,7 +212,7 @@ impl AcpTerminal {
         let output = if self.replayed_output.is_empty() {
             captured
         } else {
-            let mut out = self.replayed_output.text();
+            let mut out = self.replayed_output.text().to_string();
             out.push_str(&captured);
             out
         };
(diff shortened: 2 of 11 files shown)


ahammadnafiz:
## CI is red, and none of it is this PR

`0.3.1` is already red. This PR's failing set is a strict **subset** of the base branch's:

| | base `0.3.1` @ `b701f407` | this PR |
| -- | -- | -- |
| `atlas-agent-delta` | fail (Clippy) | fail (Clippy) |
| `atlas-agent-manager` | fail | fail |
| `atlas-agent-servers` | fail | fail |
| `atlas-agent-store` | fail | fail |
| `atlas-checkpoint` | fail | fail |
| `frontend` | fail | fail |
| `atlas-native-agent` | fail | still running |

The two crates this PR changes most both **pass**: `atlas-acp-thread` and `atlas-terminal`.

**`atlas-agent-delta` fails at the same step, for the same reason, on both.** It is `clippy::question_mark` at `crates/atlas-agent-servers/src/debug_log.rs:103` — a crate this PR does not touch — propagating through `-D warnings` to every dependent. That is ATL-234, deliberately left alone in #215 and here, so as not to half-answer it. Locally, `atlas-terminal` and `atlas-acp-thread` pass `-D warnings`, and no clippy diagnostic points at any file in this PR.

**The frontend failure is a different mechanism and may deserve attention.** `tests/vendor-licensing.test.ts > finds the modification set (parser health)` fails with `expected 0 to be greater than 5` — but only in CI. It passes locally (7/7), and the whole suite is 1031/1031 green here. The test looks for modified vendored files, so a shallow CI clone finding zero is the obvious suspect. It has been failing on `0.3.1` since before this branch and has no ticket that I can find.

Neither is mine to fix in this PR, but the second one is worth someone's five minutes — a test that only fails in CI is a test nobody can act on.

PR #225, opened 2026-09-02
fix(agent-servers): merge PATH extras with the platform separator

kevin9327:
### What?

`crates/atlas-agent-servers/src/host_env.rs` — the boot-time PATH enrichment split and rejoined `PATH` on `':'`. That is only the separator on Unix. This switches the merge to `std::env::split_paths` / `join_paths`, extracts it into a pure `prepend_entries` with tests, and skips the two parts of the enrichment that are Unix by construction (the Homebrew/`/usr` guesses and the `$SHELL -lic` probe) on other platforms.

### Why?

On Windows the separator is `;` and every entry carries a drive-letter colon, so `C:\Windows\system32;C:\Windows` came apart as `C`, `\Windows\system32;C`, `\Windows`. Two consequences:

- the dedupe never matched anything (no fragment equals a real directory), and
- the `':'` rejoin glued every extra onto the *first* inherited entry: `C:\Users\me\.bun\bin:C:\Windows\system32;C:\Windows`. Windows then reads the first entry as one directory that doesn't exist, so whatever was first on the user's PATH silently stops resolving for the rest of the process — and every ACP agent spawned after boot inherits that PATH (`AcpConnection::stdio` passes the process env through).

CONTRIBUTING names "PATH resolution" as the first thing Windows testing should look at; this is the boot-time half of it.

### How?

- `prepend_entries(base: &OsStr, extras: &[String]) -> OsString`: `split_paths` → dedupe as `PathBuf`s → `join_paths`. Empty/unset PATH is treated as no entries (it otherwise splits into one empty entry and rejoins with a trailing separator). If `join_paths` refuses (an inherited entry containing the separator — nothing we add does), the PATH is left as it was rather than lost.
- `prepend_to_path` now reads `var_os("PATH")` and just applies the result; the `SAFETY` note on the `set_var` stays where it was.
- `apply_cheap_path_extras`: the `/opt/homebrew/*`, `/usr/local/*`, `/usr/bin`, `/bin` guesses only on `cfg!(unix)` — on Windows they exist nowhere. The `$HOME`-relative ones stay (Git Bash / MSYS users do have `HOME`).
- `merge_login_shell_path`: early return unless `cfg!(unix)` — there is no `$SHELL` on Windows, the timeout path shells out to `kill -9`, and a GUI process there already inherits the user's full PATH.

No change in behaviour on macOS/Linux beyond the merge going through `split_paths`/`join_paths`, which produce the same `':'`-joined string there.

**Verification** (Windows 11, `cargo 1.95`; CI uses current stable):

Four tests in `host_env::tests`. Run against the *old* logic (temporarily lifted verbatim into `prepend_entries`) they fail on Windows with the exact shape described above:

```
test host_env::tests::extras_go_in_front_and_every_inherited_entry_survives ... FAILED
  left: ["C:\Users\me\.bun\bin:C:\Windows\system32", "C:\Windows"]
 right: ["C:\Users\me\.bun\bin", "C:\Windows\system32", "C:\Windows"]
test host_env::tests::an_extra_already_on_path_is_not_added_again ... FAILED
  left: ["C:\Windows:C:\Windows\system32", "C:\Windows"]
 right: ["C:\Windows\system32", "C:\Windows"]
test host_env::tests::the_first_listed_extra_wins ... FAILED
test result: FAILED. 1 passed; 3 failed
```

With the fix: `cargo test --locked -p atlas-agent-servers --lib` → 16 passed (12 existing + 4 new). On Unix the same tests exercise the `':'` form and pass before and after — the Windows failure is the one this change is for, so the honest statement is: the tests prove the regression only on a Windows runner.

`cargo clippy --locked --all-targets -- -D warnings` (this crate is `clippy: true` in CI) is clean for `src/`. It reports one pre-existing `collapsible_match` in `tests/connect.rs:540` under clippy 1.95 that CI's 1.98 does not flag; not touched here. `rustfmt --check` clean on the file.

## Checklist

- [x] New behaviour has a test; a bug fix has a test that fails without it — see above; fails on Windows only, by the nature of the bug
- [ ] You've run the app and used the change in a window — the app does not currently build on Windows (`numkong` does not compile under MSVC; separate report), so this is verified at the crate level only

### DIFF
--- a/crates/atlas-agent-servers/src/host_env.rs
+++ b/crates/atlas-agent-servers/src/host_env.rs
@@ -4,6 +4,9 @@
 //! It belongs with the transport because it exists entirely for the benefit of
 //! the child process an [`AcpConnection`](crate::AcpConnection) spawns.
 
+use std::ffi::{OsStr, OsString};
+use std::path::PathBuf;
+
 /// Startup-time host environment fix-ups for the ACP agent process.
 ///
 /// Two concrete problems this addresses:
@@ -62,6 +65,12 @@ fn enrich_path() {
 /// init, etc.) can't hang app startup — on timeout we fall back to the
 /// hardcoded extras already applied in `apply_cheap_path_extras`.
 fn merge_login_shell_path() {
+    // A login-shell probe is a Unix idea: Windows has no `$SHELL` (and no
+    // `kill` for the timeout path below), and a GUI process there already
+    // inherits the user's full PATH.
+    if !cfg!(unix) {
+        return;
+    }
     let shell = std::env::var("SHELL").unwrap_or_else(|_| "/bin/zsh".to_string());
 
     // `probe_shell` runs `-lic` — login AND interactive, so both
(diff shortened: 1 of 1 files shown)


pacifio:
@Ukaykhingmarma28 can you help with this review, thanks!

PR #237, opened 2026-09-04
feat(chat): pace native-agent streaming with an adaptive per-session smoother

Ukaykhingmarma28:
## What

Native-agent (Atlas Agent) responses stream into the chat in visible lumps. A live wire probe against `ai.tryatlas.cc` confirmed the cause: Claude-via-Vertex emits text in **~400–650 ms bursts** (occasional 1–1.5 s stalls), and every hop of our pipeline — gateway, engine SSE decoder, native-agent pump, Tauri emit, RAF batch — faithfully forwards per-delta, so the bursts reach the UI intact. (Full code-read + measurements in `research/stream-smoothing.md`, repo-local.)

This adds a presentation-only `StreamSmoother` between the `atlas:agents` listener and the RAF batcher in `App.tsx`:

- Per-session char queues drained on a 33 ms tick with an **elapsed-time budget**, so WebKit-throttled timers (unfocused window) degrade to chunkier-but-current, never a growing backlog.
- **Adaptive ratchet-up rate** targeting ~450 ms lag (derived from the measured burst gap), with a floor rate so an already-smooth stream passes through at chunk speed.
- **Wire order preserved**: every non-text delta becomes a queue barrier behind the text ahead of it; `turn_finished` barriers too, so the tail keeps typing before the idle flip.
- **Stop feels like stop**: cancelled turns, `turn_failed`, and `agent_disconnected` flush synchronously. Permission deltas bypass entirely (modal latency).
- Scope-gated to native-agent sessions; ACP sessions and unknown sessions pass through untouched. The persisted transcript never sees the pacing.

## Tests

12 unit tests (`stream-smoother.test.ts`, fake timers): burst spreading, elapsed-time drain under throttling, floor-rate passthrough, barrier ordering, tail-after-finish, instant cancel/fail flush, per-session independence, surrogate-pair safety.

`bun run typecheck` clean; the 3 failing suites on this branch (`atlas-comms` CI-coverage/opt-level guards + a comms midnight test) pre-date this change and are untouched by it.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,6 +12,7 @@ import { useTerminalStore } from "@/features/terminal/stores/terminal-store";
 import { useProjectStore, type AppStateWire } from "@/features/project/stores/project-store";
 import { useChatStore } from "@/features/chat/stores/chat-store";
 import { listenAgents, resetAgentByAgentId } from "@/features/chat/lib/agents-api";
+import { StreamSmoother } from "@/features/chat/lib/stream-smoother";
 import type { PendingPermission } from "@/types/acp";
 import type { AgentDelta } from "@/types/agents";
 import { cycleChatAgent } from "@/features/chat/lib/switch-agent";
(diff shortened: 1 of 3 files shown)


pacifio:
I think this is already fixed so closing this as of now, recent changes are in 0.3.2

PR #246, opened 2026-09-06
fix(chat): stop the streaming transcript flickering

monzim:
### What?

Fixes the flickering, jumping and unreadable rendering of a streaming answer in the chat transcript. Six separate causes in the chat markdown path, all in `src/lib/markdown-*` and `src/features/chat/components/`:

1. **Large tails stopped updating in dev.** `TransientMarkdown` set an `alive` ref to `false` in an effect cleanup and never set it back to `true`. StrictMode's mount/unmount/mount trips that on the first render, so every block over 2 KB went mute mid-answer and only caught up when the turn settled. The latch is now set on each effect run.
2. **Paragraph spacing changed when a turn finished.** The streaming render wraps each top-level block in a `display: contents` container, and that container carried `.atlas-prose` — so `.atlas-prose > :first-child { margin-top: 0 }` matched *every* block's first element. Inter-block gap was 0 px while streaming and 12 px the instant it settled: the whole answer re-flowed as it finished. The wrapper is now `.atlas-md-block`, with two rules restoring the same rhythm one level deeper, so streaming and settled prose are pixel-identical.
3. **The settle seam flashed raw markdown.** When a block stops being the tail it re-renders through `CachedMarkdown` with a source nothing has parsed, so anything over the sync limit fell back to its raw-source placeholder for a worker round-trip — the paragraph you were reading turned back into literal asterisks, then snapped. The tail renderer now stashes its last html (`noteTailHtml`) and `CachedMarkdown` shows that while the real parse runs. `ProseRowView` also keeps the block renderer for a row that streamed instead of switching back to a whole-message parse at the end of the turn (which re-parsed the entire answer under an uncached source).
4. **Half-written markup snapped.** `**bol` showed its asterisks for a few frames and then the words jumped to bold; `[label](htt` showed its brackets and then collapsed. The new `closeIncompleteMarkdown` repairs the tail before parsing it — dangling emphasis/strikethrough/code spans are closed, a partial link stays hidden until complete. Live tail only; settled blocks always render from raw source, so nothing repaired ever reaches the cache.
5. **Every frame replaced the DOM.** `dangerouslySetInnerHTML` is a teardown, not an update, so the nodes you were looking at stopped existing several times a second: **you could not select text while the agent was writing**, hover reset, and WebKit repainted the whole block rather than the one line that changed. Rendered html is now patched in place.
6. **Per-frame cost scaled with answer length.** The block split re-parsed the whole message every frame, so a long turn got progressively jankier as it went. While the tail is only growing it is now a substring (`splitBlocks` reports the tail offset); a real re-split runs only when the delta could have opened a new block, with a 500 ms backstop.

Plus: the streaming caret is drawn from the tail block's last element instead of a block-level `<span>` after it, so it sits after the final character rather than on a line of its own (which also removes a line of height appearing and vanishing as blocks settle).

### Why?

Reading a streaming answer was the worst-looking surface in the app: text alternating between raw and formatted, spacing shifting under you at the end of every turn, a caret hanging below the paragraph, and no way to select text until the agent stopped. Individually each cause is small; together they read as "the renderer is broken".

**This adds one top-level dependency: [`morphdom`](https://github.com/patrick-steele-idem/morphdom) (2.7.8, ~5 KB minified, zero dependencies, MIT).** Flagging it per CONTRIBUTING's "new heavy dependencies need discussion". It is the primitive for cause 5 — it walks the existing DOM against the new html and mutates only what differs (usually a single text node), which is exactly the streaming-update problem and is what Phoenix LiveView uses for it. It is confined to `src/lib/dom-html.ts`; nothing else imports it, and it does not enter the markdown worker chunk. Writing it by hand was the alternative, at ~100 lines of DOM-diffing nobody would want to own. Happy to drop it if you'd rather not take the dependency — cause 5 would then need its own in-repo patcher.

The `closeIncompleteMarkdown` rules are the same ones [Streamdown](https://streamdown.ai)'s `parseIncompleteMarkdown` applies. Streamdown itself is not usable here: it is a react-markdown renderer, and this repo parses in a Web Worker and injects sanitized html, so the rules are in-repo instead of the library.

No telemetry changes.

### How?

| File | Change |
|---|---|
| `src/lib/dom-html.ts` *(new)* | `applyHtml` — morphdom patch with a first-fill fast path; preserves the injected code-copy bar. `escapeHtml` for the raw-source placeholder. |
| `src/lib/markdown-stream.ts` *(new)* | `closeIncompleteMarkdown` — tail repair. Declines to touch anything inside an unclosed fence; leaves `_` alone (snake_case is more common in agent output than underscore italics). |
| `src/lib/markdown-blocks.ts` | `splitBlocks` (split + tail offset) and `mayStartNewBlock`; `splitTopLevelBlocks` now delegates. |
| `src/lib/markdown-cache.tsx` | Last-good-tail stash (`noteTailHtml` / `tailFallback`, prefix match with a 75 % coverage floor so a fallback can't cause a height jump); `CachedMarkdown` writes children through `applyHtml`. |
| `src/features/chat/components/streaming-markdown.tsx` | `alive` fix, tail repair, in-place patching, incremental split, adaptive inline-parse budget, `.atlas-md-block` / `.atlas-stream-tail` wrappers. |
| `src/features/chat/components/transcript-rows.tsx` | A row that streamed keeps the block renderer. |
| `src/styles/globals.css` | `.atlas-md-block` rhythm rules, inline caret, reduced-motion coverage. |

Correctness note on the incremental split: getting `mayStartNewBlock` wrong cannot break rendering. The trailing block is rendered by parsing it as markdown, so a tail that briefly holds two blocks looks identical — only cache granularity is affected, and the backstop re-split repairs that.

### Tests

New: `src/lib/markdown-stream.test.ts`, `src/lib/markdown-blocks.test.ts`, `src/lib/dom-html.test.ts`, `src/features/chat/components/streaming-markdown.test.tsx` (37 assertions). The ones that fail without this change:

- `closeIncompleteMarkdown` renders `**bold`, `**bold*` and `**bold**` as the same bold word — the frames a stream actually produces.
- `applyHtml` keeps the element identity of an unchanged sibling and of a paragraph whose text grew.
- `StreamingMarkdown` grows the tail's `<p>` in place across a rerender rather than rebuilding it, and formats `Hello **bol` as bold immediately.
- `tailFallback` lends html to a longer source that starts with it, and refuses one it barely covers.

`bun run lint`, `format:check`, `typecheck`, `test` (1196 passed) and `build` all pass; the suite was run repeatedly to confirm the new tests aren't timing-flaky.

> **Branch note:** this is the same commit as #245 (which targets `0.3.1`), cherry-picked onto `main` so the diff is the fix alone with none of `0.3.1`'s other commits. Opened against `main` on request — merge whichever fits your release flow and close the other.

## Checklist

- [ ] Targets the current version branch — this one targets `main`; see the branch note above (#245 is the `0.3.1` version-branch copy)
- [ ] **Not done — please check this before merging.** I don't have a macOS window to run the bundle in, so every claim here is from code and the test suite, not from watching a real stream. The three things worth eyeballing: paragraph spacing must not shift when a turn ends, the caret must sit at the end of the last line, and text must be selectable mid-answer.

https://claude.ai/code/session_0195fCgt87j14eUeLwiyzY5S

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -69,6 +69,7 @@
         "lucide-react": "^1.40.0",
         "matter-js": "^0.20.0",
         "mermaid": "^11.17.2",
+        "morphdom": "^2.7.8",
         "pdfjs-dist": "5.4.296",
         "pixi.js": "^8.20.1",
         "posthog-js": "^1.426.3",
@@ -1605,6 +1606,8 @@
     "min-indent": ["min-indent@1.0.1", "", {}, "sha512-I9jwMn07Sy/IwOj3zVkVik2JTvgpaykDZEigL6Rx6N9LbMywwUSMtxET+7lVoDLLd3O3IXwJwvuuns8UB/HeAg=="],
 
     "mixin-deep": ["mixin-deep@1.3.2", "", { "dependencies": { "for-in": "^1.0.2", "is-extendable": "^1.0.1" } }, "sha512-WRoDn//mXBiJ1H40rqa3vH0toePwSsGb45iInWlTySa+Uu4k3tYUSxa2v1KqAiLtvlrSzaExqS1gtk96A9zvEA=="],
+
+    "morphdom": ["morphdom@2.7.8", "", {}, "sha512-D/fR4xgGUyVRbdMGU6Nejea1RFzYxYtyurG4Fbv2Fi/daKlWKuXGLOdXtl+3eIwL110cI2hz1ZojGICjjFLgTg=="],
 
     "motion": ["motion@12.43.0", "", { "dependencies": { "framer-motion": "^12.43.0", "tslib": "^2.4.0" }, "peerDependencies": { "@emotion/is-prop-valid": "*", "react": "^18.0.0 || ^19.0.0", "react-dom": "^18.0.0 || ^19.0.0" }, "optionalPeers": ["@emotion/is-prop-valid", "react", "react-dom"] }, "sha512-BQgQbSa9Hn3/mtbib0MK53y6JSANa+YKUKlaYnWzAVDH424RYQ5LVpV3pNiWH00BA2z4ojsSdMzqT7g2FQwjuQ=="],
 
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "lucide-react": "^1.40.0",
     "matter-js": "^0.20.0",
     "mermaid": "^11.17.2",
+    "morphdom": "^2.7.8",
     "pdfjs-dist": "5.4.296",
     "pixi.js": "^8.20.1",
     "posthog-js": "^1.426.3",
--- a/src/features/chat/components/streaming-markdown.test.tsx
+++ b/src/features/chat/components/streaming-markdown.test.tsx
@@ -1,0 +1,73 @@
+// @vitest-environment happy-dom
+import { describe, it, expect, afterEach } from "vitest";
+import { render, cleanup, waitFor } from "@testing-library/react";
+import { StreamingMarkdown } from "./streaming-markdown";
+import { noteTailHtml, tailFallback } from "@/lib/markdown-cache";
+
+afterEach(cleanup);
+
+/** The block split is rAF-coalesced, so the tail only reaches the DOM a frame
+ *  after the source changes. Poll for the text rather than racing the clock. */
+async function tailReads(container: HTMLElement, text: string): Promise<void> {
+  await waitFor(() =>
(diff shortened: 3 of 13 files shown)


pacifio:
Apply it for 0.3.1, not main

monzim:
Wrong target — the fix belongs on the version branch. Superseded by #245, which is the same commit against 0.3.1.

PR #249, opened 2026-09-07
Copy and retry actions under user messages

uzayer:
Two icon buttons under each user message in the agent chat panel, like the Claude app. Copy on every message; retry on the last one, where the agent can actually rewind.

## Retry is a rewind, not a resend

Appending the same prompt to the bottom of the thread isn't a retry — it's the user re-typing, and it leaves the failed turn in the model's context for the second attempt to trip over. `EngineConnection::rewind_last_turn` drops the turn via `thread/rollback` and hands back the prompt that started it; the caller re-sends through the ordinary send path.

Rewind and send are two calls on purpose: they fail apart. A rewind that lands followed by a send that doesn't must leave the user holding their prompt, not a silently shortened thread. Every bail-out after a landed rewind hands the prompt back through `enqueueMessage`, so it shows as a composer chip the user can edit or send.

ACP has no verb for forgetting a turn — session capabilities in schema 1.5.0 are `list`/`delete`/`fork`/`resume`/`close`/`additional_directories`/`meta`. So the affordance is hidden for ACP agents rather than faked. That's asked through a new `AgentSessionRewind` seam plus `AgentConnection::supports_rewind`, alongside the existing `AgentSessionTruncate`/`AgentSessionRetry` traits, and published onto the catalog entry next to `supportsFork`. Nothing outside the native crate names a concrete connection type — an agent that grows a rewind gets the affordance for free.

## Performance

The transcript isn't virtualized (deliberately — see `transcript.tsx:1`), so this mounts for every user message in the window. The reveal is therefore pure CSS `group-hover` against the row wrapper's existing `group` class, plus `focus-within` for keyboard: a JS hover state would fire a `setState` for every bubble the pointer crosses during a fast flick, which is exactly the work the transcript is built to avoid.

The bar is absolutely positioned into the `pb-5` gap the user column already reserves, so nothing grows on hover and no row below reflows mid-scroll (house rule 1). `UserRowView`'s new props are primitives, leaving the `memo` and the projection's structural sharing intact.

## Also fixes

`history_rewound` only spliced the message log. It left `userMessageCount` permanently high and the discarded turn's plan card on screen. Pre-existing under `/undo`, but visible under retry, which rewinds and re-adds on every press.

## Known limits

- Retry is last-turn only. Arbitrary-position retry needs `ClientUserMessageId` plumbed end to end, an `AgentSessionTruncate` impl, and fork-and-replay for ACP agents.
- Replay is not byte-identical. The thread flattens content blocks, so a resource link comes back as a bare URI and no `resourceLinks` are re-sent. Pure prose round-trips exactly. Attachments are not reconstructed.
- No filesystem rollback: a rewound turn's file edits stay on disk. True of `/undo` today too.
- Retry is not serialized against queue draining.

## Verification

`format:check`, `typecheck`, `lint`, `test` (1236 passed), `cargo check --workspace`, `cargo clippy` (no new warnings in touched files), `cargo test -p atlas-native-agent -p atlas-acp-thread -p atlas-agent-delta`.

15 tests across `retry-gate.test.ts` and `retry-turn.test.ts`. The concurrency and timeout guards were verified load-bearing by neutralising them and watching the tests fail.

Not yet run in a real app window.

### DIFF
--- a/crates/atlas-acp-thread/src/connection.rs
+++ b/crates/atlas-acp-thread/src/connection.rs
@@ -220,6 +220,15 @@ pub trait AgentConnection: Send + Sync {
         false
     }
 
+    /// Whether this agent can drop its own last turn — see [`AgentSessionRewind`].
+    ///
+    /// Session-independent, like [`Self::supports_logout`], because the
+    /// catalogue publishes it per agent before any session exists. It must
+    /// agree with [`Self::rewind`]: this answers "would that return `Some`".
+    fn supports_rewind(&self) -> bool {
+        false
+    }
+
     /// Load an existing session by ID.
     fn load_session(
         self: Arc<Self>,
@@ -311,6 +320,10 @@ pub trait AgentConnection: Send + Sync {
         None
     }
 
+    fn rewind(&self, _session_id: &acp::SessionId) -> Option<Arc<dyn AgentSessionRewind>> {
+        None
+    }
+
     fn set_title(&self, _session_id: &acp::SessionId) -> Option<Arc<dyn AgentSessionSetTitle>> {
         None
     }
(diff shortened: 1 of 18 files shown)


uzayer:
One commit added that is unrelated to the retry/copy feature, because CI on `0.3.1` is red and this branch inherits it.

**`fix(clippy)`** — `app (src-tauri)` has failed since `57e8fa63`, six commits back. Two `redundant_closure_for_method_calls` sites (`auth/core.rs:279` from `57e8fa63`, `commands/keybindings.rs:294` from `edb7600a`), which the root `Cargo.toml` marks `deny`, so they are build errors rather than warnings. They went unnoticed because clippy is in no command a contributor is told to run: `bun run test:rust` runs `cargo test` alone, and `cargo check` never loads the clippy half of the lint table. So the tree passes every documented gate and still fails CI.

### Still red, not fixed here

`atlas-agent-servers::a_dying_agent_reports_what_it_said_on_stderr` fails on real `ubuntu-latest` (run 34107051588). Diagnosed but deliberately left for a separate change:

`trailing_stderr()` (`debug_log.rs:355`) scans backwards with `.rev().take_while(is Stderr)`, stopping at the first non-stderr entry. Outgoing RPC is recorded *before* the write (`connection.rs:183`), so `initialize` can land after the agent’s stderr line:

```
record stderr -> record outgoing initialize -> child exits -> trailing_stderr() == None
```

That produces the observed empty string with a fully drained pipe — the test’s `sleep 0.2` comes after its only stderr write, so it does not prevent `initialize` from becoming the final entry.

The suffix-only behaviour is deliberate: `tests/units.rs:219` (`no_trailing_stderr_when_the_last_thing_was_traffic`) pins it. It also predates #247 — that PR only added `.map(|retained| &retained.message)` and the test that exposes the bug, not the bug itself.

A fix needs an exit-path accessor that collects stderr **across** intervening RPC entries, preserving `trailing_stderr()` for its existing callers. Synchronising the stderr drain alone does **not** fix it: a log ending in `initialize` still returns `None`.

Also worth noting, it is platform-dependent — green on macOS and on arm64 Linux, red on x86_64 Linux — so it reproduces only where CI runs.

PR #264, opened 2026-09-16
Fix three Windows bugs: PATH casing, MSVC UTF-8 sources, and a first run with no Organisation

sijingl2019:
Three independent bugs found while getting Atlas to run on Windows 10 (Chinese locale, cp936). Each one is reproduced and verified; two of them are invisible on macOS by construction, which is why I've noted how to see them.

---

### 1. `with-posthog-env.mjs` drops the inherited `PATH` on Windows

`env` is a spread of `process.env` — a plain object whose PATH key is spelled however the platform spells it. Windows spells it `Path`, where `env.PATH` reads `undefined`, so

```js
env.PATH = `${binDir}${delimiter}${env.PATH ?? ""}`;
```

did not prepend to PATH. It created a **second** key holding only `node_modules/.bin` and left the real `Path` untouched. Which of the two duplicates libuv hands the child depends on the rest of the environment, so the same command worked in one shell and failed in the next:

```
$ bun run dev:app
[with-posthog-env] PostHog key not set — telemetry inert
error: bun is not installed in %PATH%          # the tauri shim, with an empty PATH
'node' is not recognized as an internal or external command
```

Fixed by reading and writing through whichever case is already present.

`tests/with-posthog-env-path.test.ts` forces both spellings on every platform, so the macOS/CI run covers the Windows path too. On the old code it fails for `Path` and passes for `PATH` — exactly the split that made this look nondeterministic.

---

### 2. MSVC reads UTF-8 sources in the system ANSI codepage

On a Chinese or Japanese Windows (cp936 / cp932) a double-byte lead byte inside a UTF-8 comment pairs with the byte after it and swallows the newline, so the next line of code disappears into the comment. `numkong` (via `usearch` → `atlas-memory`) has comments containing real `∞`, `→` and `≤` characters, and on cp936 they broke `cast/serial.h` with a cascade starting at

```
include\numkong/cast/serial.h(884): error C2059: syntax error: 'if'
```

Adding `/utf-8` makes the identical `cl.exe` command exit 0. Target-scoped in `.cargo/config.toml`, so macOS builds are unaffected; cc-rs appends it to each crate's own flags rather than replacing them.

This is latent for every CJK-locale Windows contributor and cannot reproduce on an English install.

---

### 3. A first run comes up with zero Organisations, and "Open Folder" silently does nothing

`AppState::load` returned `Self::default()` directly on its two early-return paths — no resolvable app data dir, or **no `state.json` yet** — which skipped `migrate()`. A fresh install therefore had `organisations: []`.

Every render surface filters strictly by the active org, and `addWorkspace` refuses outright without one:

```ts
const org = requireActiveOrgId();
if (!org) { logEvent({ kind: "workspace-add-refused", ... }); return null; }
```

So on a brand-new install the folder picker opens, you choose a directory, and nothing happens at all — no error, no toast, just a log line. Meanwhile `project-store.ts` documents the contract this broke:

> Rust `migrate()` guarantees a default "Personal" org + an `activeOrganisationId` … so this is always set.

That held on every path except the first launch.

Confirmed from one install's `state.json` timeline:

| time | event |
|---|---|
| 13:36 | first launch, no `state.json` → `migrate()` skipped → `organisations: []` → Open Folder dead |
| 13:46 | relaunch, file now exists → `migrate()` runs → `"Personal"` created → project opens |

The self-healing second launch is probably why this survived: it only bites the very first impression.

The parse+migrate half of `load` is split into `from_raw`, reachable without an `AppHandle`, and `None` (first run) now migrates like any other input. `a_first_run_still_gets_the_default_organisation` covers it.

---

### Verification

| | |
|---|---|
| `bun run typecheck` | passes |
| `vitest run tests/with-posthog-env-path.test.ts` | 2 passed; fails on the old code for the `Path` spelling |
| `/utf-8` | same `cl.exe` command goes from a wall of C2059 to exit 0 |
| `cargo build -p atlas --no-default-features` | passes; `tauri dev` reaches `Running DevCommand` and the app launches |

One gap I should flag rather than paper over: **I could not execute the Rust unit test on this machine.** `cargo test -p atlas --lib` fails to *load* its 238 MB test binary with `STATUS_ENTRYPOINT_NOT_FOUND (0xc0000139)`, while `cargo test -p atlas-git --lib` in the same workspace runs 28 tests fine. Ruled out: the VC++ runtime (system 14.51 is newer than the 14.44 toolset), the 95 proc-macro DLLs in `target/debug/deps` (a copy in a clean directory fails identically), delay-load imports (none), and binary size (fails with debuginfo off too). Its static imports are a subset of the `atlas.exe` that loads fine, plus one valid `kernel32` export. This looks like a pre-existing Windows-only issue with that particular test binary and is unrelated to the change — but it does mean `a_first_run_still_gets_the_default_organisation` has only been compiled here, not run. It should execute normally on macOS/CI.


Ukaykhingmarma28:
Thanks — the third finding is exactly right, and it is the first-launch "Open Folder does nothing" bug several users have reported on both macOS and Windows. We reproduced the mechanism on `0.3.3`:

- `AppState::load` returns `Self::default()` from its two early returns without `migrate()`, so a fresh install boots with zero organisations.
- A few lines later in `setup()` the telemetry-id / config-marker save writes `state.json` with `organisations: []`, which is why the second launch heals itself.
- The frontend's `requireActiveOrgId()` then refuses in `addProject`, silently.

The early return predates orgs; it became user-visible with b7d0c038 (2026-08-31), which made the org filters strict and the refusal hard. So it affects every 0.3.1+ fresh install that is not signed in.

The `app_state.rs` change from this PR is landing on `0.3.3` in #279 with attribution, plus a toast on the refusal branch so it can never be silent again. Your new test runs on macOS and passes. Could you split the other two hunks (the `Path` casing in `with-posthog-env.mjs` + its test, and the MSVC `/utf-8` flag) into their own PR? They look plausible but need separate review, and the org fix should not wait on them.

PR #270, opened 2026-09-18
macOS: DMG Finder branding and Liquid Glass app icon

zuhayermasud:
## Summary
- Adds a custom DMG Finder background, window layout, and volume/drive icon for the release `.dmg`, driven by a new `scripts/layout-dmg.sh` (hand-rolled Finder AppleScript styling, since `release-macos.sh` builds its dmg with a bare `hdiutil create` rather than Tauri's own bundler).
- Adds Icon Composer sources (Default/dark and Light styles) for the macOS 26 Liquid Glass app icon, and wires the Default style into `tauri.conf.json`'s `bundle.icon` so `tauri build` compiles an `Assets.car` via `actool` alongside the `.icns` fallback (renamed `icon.icns` → `Icon.icns` to match the name `tauri-bundler` derives from the compiled asset catalog).
- Two Finder-quirk fixes for the DMG layout script, found while verifying the preview build:
  - The mounted volume's custom icon has to be applied *after* Finder's own open/close/"update disk" styling sequence — doing it earlier gets it stripped as a stale orphan.
  - The DMG window height is padded 32pt beyond the background art's design size to absorb Finder's Path Bar (a per-user preference outside the dmg's control) so the background doesn't clip for anyone who has it enabled.

## Test plan
- [x] `bun run typecheck:app` / `typecheck:test` and `vitest run tests/` pass (via pre-commit hooks on each commit)
- [x] Built local preview `.dmg`s against a staging `Atlas.app` and visually verified: background fills the window edge-to-edge, icon/Applications positions, and the custom volume icon shows in the titlebar/path bar/Desktop, with Finder's Path Bar both on and off
- [ ] Full `tauri build --bundles app,dmg` with the Liquid Glass `.icon` — blocked locally pending Xcode (only Command Line Tools installed; `actool` requires full Xcode.app)


uzayer:
Merged into 0.3.3 — thank you, @zuhayermasud! The DMG Finder branding and the Liquid Glass icon both look great, and the two follow-up fixes (stamping the volume icon after Finder styling, and padding the window height for the Path Bar) were exactly the right catches.

I landed this with three small commits on top rather than asking for another round:

- `339eb086` swap the file and volume icons back
- `475ef3c3` compile the icon with Xcode's `actool` so the Liquid Glass variants come through
- `cd6e4d78` a setting to pick the dark or light icon

All of it is on `0.3.3` now. Appreciate the work.

PR #278, opened 2026-09-20
feat(linux): port Atlas to Linux with packaging pipeline and desktop integration

ik0zy:
### What?

Ports Atlas to Linux across modern desktop environments (GNOME, KDE Plasma, Hyprland / Wayland / X11) with native window controls, desktop integration, keybindings, shell fallbacks, and an automated release workflow for `.deb`, `.rpm`, `.AppImage`, standalone `.tar.gz`, and Arch Linux `.pkg.tar.zst` packages.

### Why?

Linux is highlighted as an active area of contribution in `CONTRIBUTING.md` ("*Linux and Windows testing of the production bundle — terminal font, PATH resolution, general GUI behaviour*"). Currently, running Atlas on Linux encounters several platform gaps:
- **Duplicate GTK Window Frame:** On Wayland/GNOME/Hyprland, the default server-side window frame duplicates the web titlebar.
- **Missing Window Controls:** The web titlebar renders macOS traffic lights or Windows minimize/maximize/close buttons, but has no Linux handling.
- **Shell Fallback & Sudo Relaunch:** Assumes `/bin/zsh` when `$SHELL` is empty or unset, and sudo re-executes `zsh -c` even on systems where bash is the default shell.
- **Terminal Opening:** "Open in Terminal" in the file explorer only implements macOS `open -a Terminal` and fails on Linux.
- **CLI Helper Overwriting:** `cli_install_helper` and `App.tsx` can overwrite a compiled ELF binary at `~/.local/bin/atlas` with the shell wrapper and trigger an infinite recursion loop.
- **No Linux Packaging Pipeline:** No automated release bundles (`.deb`, `.rpm`, `.AppImage`, `.pkg.tar.zst`, or standalone tarball) exist in CI.

*Note on dependencies & telemetry:* Adds zero new JavaScript or Rust production dependencies. Telemetry pipeline is untouched.

### How?

- **Window Management & Client-Side Decorations:**
  - Added `src-tauri/tauri.linux.conf.json` configuring `"decorations": false` to avoid double titlebars on Wayland.
  - Updated `src/lib/platform.ts` to export `isLinux = ua.includes("Linux")`.
  - Updated `src/components/titlebar.tsx` to render `<WindowControls />` on Linux with responsive header padding.
- **Keybindings:**
  - Updated `src/features/keybindings/lib/combo.ts` to display the `Super` key symbol when `Meta` and `Ctrl` are active, and added modifier parsing aliases for `super` and `win`.
  - Added unit tests in `src/features/keybindings/lib/combo-linux.test.ts`.
- **Terminal & Shell Integration:**
  - `crates/atlas-terminal/src/lib.rs` & `crates/atlas-agent-servers/src/host_env.rs`: Safe shell fallback to `/bin/bash` or `/bin/sh` when `$SHELL` is unset or empty; guarded `zsh_integration_dir()`.
  - `src/features/terminal/lib/terminal-session.ts`: Guarded `sudo` relaunch logic so non-zsh shells do not invoke `sudo zsh -c`.
  - `src-tauri/src/commands/fs.rs`: Implemented `fs_open_in_terminal` on Linux supporting modern desktop terminal emulators (`xdg-terminal-exec`, `ptyxis`, `gnome-terminal`, `kitty`, `foot`, `alacritty`, `ghostty`, `wezterm`, etc.), resolving parent directories when opening a file.
- **CLI Helper Protection:**
  - `src-tauri/src/commands/cli.rs`: Added `is_elf_binary` (`\x7fELF`) detection to prevent `cli_install_helper` from overwriting a real compiled binary at `~/.local/bin/atlas`.
  - `src-tauri/bin/atlas-cli.sh`: Hardened search paths (`/usr/bin`, `/usr/local/bin`, `/opt/atlas/bin`) and eliminated recursive self-invocations.
  - `src/App.tsx`: Guarded against re-invoking `cli_install_helper` when already installed on Linux.
- **Packaging & CI Release Pipeline:**
  - Added XDG desktop file `src-tauri/resources/dev.atlas.ide.desktop` (`StartupWMClass=Atlas`).
  - Added `scripts/package-linux.sh` to assemble standalone portable `.tar.gz` archives (with install/uninstall scripts, icons, licenses) and AUR `atlas-bin` `PKGBUILD`.
  - Added `.github/workflows/release-linux.yml` for automated release building of `.deb`, `.rpm`, `.AppImage`, standalone `.tar.gz`, and Arch Linux `.pkg.tar.zst` (via Docker), automatically attaching them to GitHub release tags (`alpha-*`, `v*`, `exp-*`).

## Checklist

Only the things CI can't check for you:

- [x] You've run the app and used the change in a window


uzayer:
Thanks @ik0zy, this is a solid Linux port, and the ELF/system-binary handling for the CLI helper is careful work. I merged it onto the tip of `0.3.3` locally and ran the frontend job, `atlas-terminal` and `atlas-agent-servers` (tests + clippy) on an x86-64 Linux box. All green. Two things need fixing before it can go in, plus a few smaller ones.

**Please retarget this PR to `0.3.3`.** Feature PRs go into the current version branch, and `0.3.3` is the one shipping next (see CONTRIBUTING.md). It merges cleanly there.

**Must fix**

1. **`src-tauri/tauri.linux.conf.json` breaks the build on Linux.** It sets `"macOSPrivateApi": false`, but `src-tauri/Cargo.toml` enables the `tauri/macos-private-api` feature, so the build script aborts with "The `tauri` dependency features … does not match the allowlist". `cargo check`/`clippy -p atlas` fail on Linux with this file present, and they pass once the line is removed. It does nothing on Linux anyway, and `tauri.windows.conf.json` doesn't set it. Please delete that line. CI runs the app crate on macOS, so it wouldn't have caught this.

2. **`.github/workflows/release-linux.yml` would overwrite existing releases.** It fires on any `alpha-*`, `v*` or `exp-*` tag and publishes with `name: Atlas <version>`, `prerelease: false` and `generate_release_notes: true`. Releases here are created by hand with their own titles and already carry the macOS/Windows assets (e.g. `alpha-0.3.3`), and `softprops/action-gh-release` would rename that release and touch its notes, and would flip `exp-*` pre-releases to full releases. Please make the workflow only attach assets: drop `name`, `generate_release_notes` and `prerelease`, and restrict the trigger to `alpha-*`. Also pass `github.event.inputs.*` through `env:` instead of interpolating them into `run:` scripts.

**Smaller things**

- `terminal-session.ts`: `Boolean(reg.zshDir)` is a no-op, since `reg.zshDir` was already truthy-checked. The real fix is the `zsh_integration_dir()` change, so you can revert this line.
- `App.tsx` no longer re-runs `cli_install_helper` on every launch when the installed version matches. That's fine on Linux, but it means a deleted or edited helper no longer self-heals on macOS. Please limit the skip to Linux (`isLinux`), or say in the description that the macOS change is intentional.
- AppImage installs get a `~/.local/bin/atlas` helper that exits with "could not find Atlas installation", because there's no `atlas` on PATH. Either skip installing the helper for AppImage (`$APPIMAGE` is set) or have it exec `$APPIMAGE`.
- Please confirm the undecorated window can be resized by dragging its edges on GNOME/Wayland and on Hyprland, and mention which environments you tested.

I haven't built the packages or run the packaging script, so that part hasn't been verified on my side. If you can also tell us how you tested the `.deb`, `.rpm` and AppImage, that would help. Once the two blockers are fixed and it's retargeted, this is good to go.

ik0zy:
All points have been addressed and the PR has been retargeted:

### 1. Retargeted to `0.3.3`
- Rebased cleanly onto the tip of `upstream/0.3.3` and changed the PR base to `0.3.3`.

### 2. Fixed `tauri.linux.conf.json`
- Removed `"macOSPrivateApi": false` so `cargo check`/`clippy -p atlas` succeed cleanly on Linux without allowlist mismatch.

### 3. Hardened `release-linux.yml`
- Restricted tag trigger strictly to `alpha-*`.
- Dropped `name`, `generate_release_notes`, and `prerelease` from `softprops/action-gh-release@v2` so it purely attaches the Linux package artifacts to existing releases without renaming them, modifying notes, or altering pre-release status.
- Passed `github.event.inputs.*` safely through `env:` variables in the version extraction step.

### 4. Reverted `terminal-session.ts`
- Reverted `Boolean(reg.zshDir)` back to `reg.zshDir`.

### 5. Preserved macOS Helper Self-Healing in `App.tsx`
- Scoped the matching-version skip check strictly to `isLinux`. On macOS, launch-time helper re-installation remains untouched so hand-deleted or edited helpers continue to self-heal.

### 6. AppImage CLI Helper Support
- Updated `src-tauri/src/commands/cli.rs` and `src-tauri/bin/atlas-cli.sh` to capture `$APPIMAGE` at install time. When Atlas is run from an AppImage, `~/.local/bin/atlas` directly executes the AppImage target. `cli_status` also validates that the referenced AppImage file exists, self-healing if the AppImage is updated or moved.

### 7. Window Resizing & Desktop Environment Verification
I have tested the undecorated window resizing on Arch Linux under both **GNOME + Wayland** and **Hyprland + Wayland**:
- **GNOME 47 / Wayland (Mutter):** The window can be resized smoothly by grabbing the border edges (~8–10px invisible grab margin provided by the compositor) as well as using `<Super> + Right-Click` drag, with no visual tearing or glitches.
- **Hyprland / Wayland:** Tiling mode automatically manages tiled window sizing as expected. In floating mode, the undecorated window can be resized cleanly by dragging its borders or using standard mouse drag (`$mainMod` + right mouse button / `bindm = $mainMod, mouse:273, resizewindow`).

### 8. Packaging Verification Details
- **Arch Linux (`.pkg.tar.zst` & AUR `PKGBUILD-atlas-bin`):** Built via containerized `makepkg`, installed locally on host Arch Linux via `pacman -U`, verified desktop menu launcher, hicolor icons (32x32 to 512x512), and confirmed CLI launch.
- **AppImage:** Built via Tauri bundler, executed on Linux x86_64, verified single-instance window reuse, and verified `~/.local/bin/atlas` execution.
- **Standalone Tarball (`.tar.gz`):** Extracted and run directly.
- **`.deb` & `.rpm`:** Verified package payloads, desktop entries, and dependencies (`libayatana-appindicator3-1` for Debian/Ubuntu, `libayatana-appindicator-gtk3` for Fedora/RHEL).

uzayer:
Merged, thanks @ik0zy! The two blockers and the smaller points are all addressed. It needed two more fixes, which I pushed on `0.3.3` in c9f0c0e0:

1. `App.tsx` read `status.installed_version` / `status.current_version`, but `CliStatus` serialises as camelCase, so both were `undefined` and the Linux "up to date" check always passed. Stale helpers were never refreshed, and a moved AppImage never triggered a reinstall. It now uses the shared `CliStatus` type.
2. `cli_status` was a sync command, so on Linux `is_atlas_binary` could scan up to 32 MiB of `/usr/bin/atlas` on the main thread at every launch. It and the checks at the start of `cli_install_helper` now run in `spawn_blocking`.

Issue #86, opened 2026-08-03
Unify slash-command picker with skills.sh skill discovery


uzayer:
## Problem

Atlas’s Skills area and the `npx skills` CLI are part of the same skills.sh ecosystem. The CLI installs skills in the shared `~/.agents/skills` store and projects them into each selected agent’s skill directory. For example:

```
~/.claude/skills/<skill> -> ../../.agents/skills/<skill>
```

Atlas correctly scans `~/.claude/skills`, but does not list skills installed by `npx skills`. Its external-skill scanner rejects every symlink before checking whether it resolves to a valid skill directory. The scanner assumes symlinks are Atlas-managed projections; skills.sh uses symlinks for normal agent projections too.

## Goal

Recognize skills.sh-installed agent projections in Atlas’s existing registry, then make the chat composer use one slash-command surface for commands and skills.

## Scope

1. Follow and safely classify valid skill-directory symlinks in agent skill directories, including skills.sh projections into `.claude/skills` and `.codex/skills`.
2. Preserve Atlas-managed projection handling and deduplicate a skill visible through multiple agent projections.
3. Do not introduce another registry, reinstall skills, or copy skills.sh-managed files.
4. Remove `#` as a chat skill-picker trigger and stop inserting `#skill:<name>` invocation mentions.
5. Trigger the `/` picker for the slash token at the caret anywhere in the composer, rather than only at the beginning of a one-line message.
6. Show both active-agent commands and discovered, applicable skills in that picker.
7. Route `/skills` to the existing skills browser/modal that `#` currently opens.

## Acceptance criteria

- A real skill at `.agents/skills/foo/SKILL.md`, projected as a symlink at `.claude/skills/foo`, appears in Atlas as an external Claude Code skill.
- The equivalent Codex projection is discovered, and the same skill is not duplicated when both projections exist.
- Atlas-managed symlink projections remain recognized as managed, without being duplicated or downgraded to external.
- Dangling symlinks, symlinks outside valid skill directories, and entries without `SKILL.md` are ignored safely.
- Typing `hello /clear` opens and filters the picker at `/clear`; selecting a result replaces only that slash token and preserves surrounding text.
- `#` has no special chat-picker behavior.
- The slash picker lists applicable commands and discovered skills.
- Typing `/skills` opens the existing skills browser/modal.
- Coverage includes skills.sh symlink discovery, deduplication/invalid links, slash detection and replacement, command/skill aggregation, and `/skills` routing.

## Related work

[...]

PR #84, opened 2026-08-02
fix: use active ACP slash commands

shyan23:
Fixes #25

The slash-command picker now uses the active ACP session's advertised commands for external agents instead of always falling back to the curated Claude catalogue.

Changes:
- Normalize malformed, duplicate, and login command records safely.
- Preserve Atlas-owned Claude Code and Codex login commands.
- Keep native Cersei behavior unchanged.
- Add no new UI elements or styling.
- Add ESLint 9 flat-config support for the TypeScript/JSX frontend.

Verification:
- `git diff --check` ✅
- `bun run lint` ✅
- `bun x tsc --noEmit` ✅
- `bun run build` ✅
- Pure command-normalization behavior reviewed ✅

Issue #81, opened 2026-08-02
Preserve agent identity and use the Atlas activity mark while a chat turn starts


uzayer:
## Problem

When a user sends the first prompt in a fresh Claude Code session, the chat panel immediately replaces the known Claude Code identity with a generic assistant/thinking treatment: a glowing, pulsing `Sparkles` icon. This makes the session appear to have defaulted to an unspecified “Assistant,” even though the tab was already created and bound as `claude-code`.

The generic indicator should not obscure agent identity. Remove the glowing star and use Atlas’s existing in-progress mark in the thread’s pre-token state.

<img width="326" height="152" alt="Image" src="https://github.com/user-attachments/assets/ee78d794-450d-43e3-84ac-2d9297446b30" />

## Investigation

* A new chat session is created with `agentType: "claude-code"` by default in `src/features/chat/stores/chat-store.ts`; the chat panel chooses the Claude ACP plugin from that stored type before binding the session.
* `ChatPanel` passes `session.agentType` into `MessagesList`, which resolves the correct per-agent avatar and label for actual assistant-message rows.
* The gap is `WorkingIndicator` in `src/features/chat/components/messages-list.tsx`: it has no agent identity input and hard-codes Lucide `Sparkles` with `animate-pulse`. This is the transient row shown after the user message while the session is running but before the first assistant token arrives.
* The Chats sidebar already renders `AtlasLoader` for active sessions in `src/features/chat/components/session-sidebar.tsx`. `AtlasLoader` is the four-bar Atlas activity mark, animated by `.atlas-loader` / `@keyframes atlas-l23` in `src/styles/globals.css`.
* The motion is an infinite **loop** with a lateral/shuffling background-position animation; it is not the generic pulse/breathing treatment.

## Desired behavior

* A Claude Code session remains visibly Claude Code as soon as it is created and throughout the first-turn wait; no UI state should imply it has become an unnamed assistant.
* Replace the in-thread glowing star entirely with the same Atlas activity mark and looping shuffle used for an in-progress chat in the Chats sidebar.
* Use the shared `AtlasLoader` component rather than duplicating its markup or animation so the two locations stay visually in sync.

## Acceptance criteria

* After sending the first prompt in a new Claude Code session, the pre-token row does not render `Sparkles`, `animate-pulse`, or a generic assistant icon.
* The row displays the shared animated Atlas activity mark with the same motion, color treatment, and reduced-motion behavior as the active-session marker in the Chats sidebar.
[...]

PR #280, opened 2026-09-21
Shared memory unification: one record store, one tool server, every agent (#77, tickets #78–#90)

Ukaykhingmarma28:
## What this is

The shared-memory unification: spec Ukaykhingmarma28/atlas#77, delivered as thirteen ordered tickets (Ukaykhingmarma28/atlas#78 through #90) plus a fourteenth step that makes memory pull-only (ADR-0010), rebased onto the current `0.3.3` line by merge. Every agent Atlas runs now reads and writes one memory store, through one seam, with provenance the user can see and edit.

## The fourteen steps

1. **Envelope** (#78): `atlas-memory` owns the injected-block envelope and every reader strips it, so injected memory can never be re-captured as a fact.
2. **Capabilities** (#79): ACP agent capabilities are plumbed to the app; HTTP MCP support is logged per agent.
3. **Record store** (#80): a SQLite record store behind the Shared tab, with a one-time migration from the memdir files and repository-level scope.
4. **Session lifecycle** (#81): session start and end are recorded, and the Shared tab refreshes live.
5. **Memory tool server** (#82): an in-process MCP server over loopback HTTP exposing the four memory tools.
6. **Every agent gets it** (#83): the tool server is handed to every agent that can take an HTTP MCP server.
7. **Briefing** (#84): a session-start briefing with a ranked index, and a floor for short prompts.
8. **Extractor** (#85): an extractor with a gateway mode replaces the legacy per-turn distill.
9. **Handoff** (#86): an agent-neutral recent-session handoff built from capture, not per-agent transcripts.
10. **Provenance** (#87): the Shared tab shows source and confidence per entry, with edit and forget.
11. **Claude import** (#88): a consented import of Claude Code's auto-memory, with a preview.
12. **Global promotion** (#89): cross-repository promotion runs over Facts in the record table; the grafeo graph, consolidation and dream modules are removed.
13. **Deletions** (#90): the superseded paths go.
14. **Pull only** (ADR-0010, commit `1175808c`): nothing is prepended to a prompt any more. `agents_send` sends the user's text as typed; the agent reads memory through the tool server, now seven tools organised as a protocol: `memory_briefing` (what the first send used to carry, sets the session clock), `memory_changes` (what other sessions wrote since), `memory_search`, `memory_get`, `memory_list`, `memory_remember`, `memory_forget`. The server's `instructions` tell the agent to call `memory_briefing` first. The native agent now waits a bounded time for its MCP servers before a turn, so the first prompt sees the memory tools; without that wait the first turn reported no MCP servers at all (reproduced on a second Mac). About 2,000 lines of push machinery are deleted.

## The merge with 0.3.3

`113d9c3b` merges `upstream/0.3.3` (theme system, Liquid Glass icons, Windows updater, OrcaRouter, first-run organisation fix, test hygiene). Nineteen conflicts; the merge commit message records each resolution. The notable ones:

- `atlas-memory`: this branch's module deletions win; upstream had only touched their tests. `tests/cersei_parity.rs` takes upstream's rename to `tests/behaviour.rs` and its tempfile hygiene, minus the graph and dream sections.
- Shared tab: the Claude-memory import dialog is ported from Radix to the Base UI dialog and from the retired tokens and `text-[Npx]` sizes onto the theme scales, so the new design-system ratchet and CSS-variable coverage suites pass.
- Upstream's workspace-to-project rename is applied to the memory code that landed after it.
- `ipc-contract` keeps both sides (dead-command allowlist plus dynamic-invoke tracking); `wire-shape` reads `EventKind` from its new home in `atlas_memory::record`.

## Verification

| | |
|---|---|
| `bun run typecheck` | passes |
| `vitest run tests/` | 18 files, 166 tests pass |
| `cargo test -p atlas-memory` | 64 unit + 15 behaviour tests pass, 4 model-gated tests ignored as upstream made them |
| `cargo test -p atlas-agent-manager` | 56 tests pass |
| `cargo check -p atlas -p atlas-native-agent --tests` | clean |
| `cargo test -p atlas --lib` | 524 tests pass, 5 ignored |

## Testing it by hand

With sharing on for the project, start a session with Atlas Agent or Claude Code and ask the agent to list its `memory_` tools, then to call `memory_briefing`. Memory ▸ Shared updates live on `memory_remember` and `memory_forget`. The offer decision is logged per session start as `memory tool server offer: agent=… http_mcp=… memory_server=included|omitted reason="…"`.


pacifio:
Pushed `aa74f3c3` to this branch to clear the clippy errors that were failing CI — five files, lints only, no behaviour change.

`scripts/test-rust.sh` notes in its own comments that it skips clippy, which is why the verification table here is green while CI is not: the tests do pass, the lints didn't.

- `atlas-memory/src/extract.rs` — `and_then(serde_json::Value::as_f64)`
- `atlas-memory/src/record.rs` — `old.key` is the last use of `old`, so the clone was dead; `search_events(..).len()` rather than collecting a `Vec` to read its length
- `atlas-memory/src/lib.rs` — three `MemoryEngine::open(root)` where the clone was the last use. The fourth, in `reuses_a_cached_vector`, is deliberately kept: `root` is opened again later in that test
- `atlas-agent-servers/src/session_mcp.rs`, `atlas-agent-manager/tests/session_mcp.rs` — a named type for the settle log in each (`type_complexity`), and `map(std::string::ToString::to_string)`

Verified per crate with CI's own invocation (`cargo clippy --locked --all-targets -- -D warnings`, run from inside the crate directory — a root-level `-p` run unifies features differently and reports a different set). `cargo test -p atlas-memory` still passes 64 + 15 with 4 ignored, matching your table.

Two lints are left alone because they only fire on Windows and are invisible to Linux CI: `atlas-agent-servers`' `pgid` field, read only under `#[cfg(unix)]`, and `atlas-agent-store`'s `needless_return` in a Windows-only branch. Both pre-date this PR.

One thing worth a re-run rather than a fix: `codex-otel`'s `otlp_http_exporter_sends_traces_to_collector` failed in the last run but not the one before it, and different jobs failed across the two — that reads as flaky rather than a regression from this branch.

